PR #72, opened 2014-10-01
Documentation: Roxygen comments

lindbrook:


jennybc:
I think the "must fix" content from here, at least, has occurred via other PRs.

PR #111, opened 2014-10-31
add note on "pander" and "asis"

daroczig:
When using the generic S3 method, the "asis" chunk option is not needed any more: http://blog.rapporter.net/2014/09/pander-tables-inside-of-knitr.html

This might be useful for someone working on the documentation, as IMHO it's not a good practice to call `pandoc.table` directly.

---

Although please feel free to close this pull request without merging if you do not think this comment is useful here.


jennybc:
I'm not sure what to do with this very, um, well-aged PR. We're resetting to work on a 2nd edition. I will just let it go quietly into the night. Thanks.

PR #196, opened 2015-01-05
small fixes to vignettes material

jennybc:
noticed a few things while I read this …

re: knitr chunk options, wondered if you also want to recommend naming chunks for later assistance when debugging and better names for figure files?


hadley:
I don't usually bother with naming - the chunk navigation in RStudio lists the numbered ones so I just use that to jump around

Issue #173, opened 2014-12-14
Define hyperlinks/anchors in tests.html


BernhardKonrad:
http://r-pkgs.had.co.nz/tests.html has the hyperlinks (under `contents`)

```
Test workflow
Test structure
Tests
Files
CRAN notes
```

but [only the first one works](http://r-pkgs.had.co.nz/tests.html#test-workflow).

PR #210, opened 2015-01-13
Add anchors to h2s so that they're not undefined in the navigation.

srushe:
I noticed that a few of the left-hand navigation links weren't working, and then I noticed issue #173 mentioning the same problem, so I added the anchors to make these work.


hadley:
Thanks - I appreciate this :)

PR #212, opened 2015-01-14
Update vignettes.rmd to explicitely mention the `build_vignettes = TRUE` option.

mattflor:
Maybe it helps to explicitely point out that you need to use `build_vignettes = TRUE` with `devtools::install()`.

When I was missing my freshly written vignette in my package after using RStudio's "build and reload", I first ran the function without the option and was still missing the vignette. Only after I had checked the other three suggestions did I take a closer look at `devtools::install()` and noticed the option...


hadley:
I'm going to change the vignette process so that vignettes results are saved in the vignettes directory. That will resolve this problem

PR #229, opened 2015-01-20
fixed two typos

tonyfischetti:
Also, the word "focussed" is used twice in the text. At first I thought this was a typo because I've never seen it spelled like that but it's correct. However, most style guides I've consulted suggest using only the single 's' variety. Should I make a pull request to s/focussed/focused/g ?

### DIFF
--- a/check.rmd
+++ b/check.rmd
@@ -6,7 +6,7 @@ output: bookdown::html_chapter
 
 # Automated checking {#check}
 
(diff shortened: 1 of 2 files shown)


hadley:
Thanks! 

Wiktionary has "The spelling focused is much more common in the US; however, the spelling focussed is sometimes used in the UK and Canada, and is especially common in Australia and New Zealand."

tonyfischetti:
Oh ok, I thought it might :)

PR #230, opened 2015-01-22
Switch anchors around on `notes`

ateucher:
`$notes^` -> `^notes$`


hadley:
Thanks! I think someone else noticed this and fixed it

PR #232, opened 2015-01-23
Indicate what to do with undocumented methods

lionel-:


hadley:
No, you just don't document them.

lionel-:
hmm I got a `check()` complaint about undocumented parameters because of that...  I may be missing something.

Ah yes, must be because I used `@rdname`. L'esprit de l'escalier. Sorry for the noise.

PR #233, opened 2015-01-25
Update check.rmd

LechMadeyski:
Dear Hadley,

It seems that instead of `devtools::install_deps(deps = TRUE)' should '`devtools::install_deps(dependencies = TRUE)' [Package devtools version 1.7.0.9000]

Regards Lech

PS Thank you for your excellent book!


hadley:
Thanks!

PR #234, opened 2015-01-27
'alternative documentation workflow' _will_ show links

benmarwick:
added 'not', seems to make more sense in context, but I'm not 100% if that's what you mean...

This seems a bit over-the-top for a one-word change, but to honor your instructions to contributors: "I assign the copyright of this contribution to Hadley Wickham"


hadley:
Thanks!

benmarwick:
No worries, thanks for making this excellent manual freely available.

PR #236, opened 2015-01-29
Fixed typos in package structure page

PeterDee:
Just some more typos I found on the train to work - not sure about the "an vs. a" ones, but changed for consistency.


hadley:
Thanks!

PR #245, opened 2015-02-02
Update how to specify an exact version of R package

jdblischak:
In DESCRIPTION file, specifying an exact version with a single `=` gives an error. Using `==` worked. I tested this using R 3.1.2.


hadley:
Good catch, thanks!

PR #247, opened 2015-02-04
mostly typo fixes

dlukes:
Just a handful of very minor fixes for your consideration :) Thank you so much for all the awesome R resources you're putting out!

### DIFF
--- a/man.rmd
+++ b/man.rmd
@@ -21,7 +21,7 @@ R provides a standard way of documenting the objects in a package: you write `.R
 * It abstracts over the differences in documenting different types of objects,
   so you need to learn fewer details.
 
-As well as generating `.Rd` files, roxygen2 can also manage your `NAMESPACE` and the `Collate` field in `DESCRIPTION`. This chapter discusses `.Rd` files and the collate field. [`NAMESPACE`](#namespace) describes how you can use roxygen2 to manage your `NAMESPACE`, and why you should care.
+As well as generating `.Rd` files, roxygen2 can also manage your `NAMESPACE` and the `Collate` field in `DESCRIPTION`. This chapter discusses `.Rd` files and the collate field. The chapter [Namespaces](#namespace) describes how you can use roxygen2 to manage your `NAMESPACE`, and why you should care.
 
 ## The documentation workflow {#man-workflow}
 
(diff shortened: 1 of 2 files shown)


man.rmd
```diff
@@ -21,7 +21,7 @@ R provides a standard way of documenting the objects in a package: you write `.R
 * It abstracts over the differences in documenting different types of objects,
   so you need to learn fewer details.
 
-As well as generating `.Rd` files, roxygen2 can also manage your `NAMESPACE` and the `Collate` field in `DESCRIPTION`. This chapter discusses `.Rd` files and the collate field. [`NAMESPACE`](#namespace) describes how you can use roxygen2 to manage your `NAMESPACE`, and why you should care.
```
dafydd-lukes: When the link is formatted as code (backticks), it's not clear from the styling that it's a hyperlink (no colour, no underlining until one hovers one's mouse over it).

package.rmd
```diff
@@ -291,6 +291,6 @@ When you use `library(pkg)` to load a package, R looks through each path in `.li
 library(blah)
 ```
 
-The main difference between `library()` and `require()` is what happens when a package isn't found. While `library()` throws an error, `require()` prints a message and returns FALSE. In practice this distinction isn't important because when building a package you should __NEVER__ use either inside a package. See [package dependencies](#dependencies) for what you should do instead.
+As stated above, the main difference between `library()` and `require()` is what happens when a package isn't found. While `library()` throws an error, `require()` prints a message and returns FALSE. In practice this distinction isn't important because when building a package you should __NEVER__ use either inside a package. See [package dependencies](#dependencies) for what you should do instead.
```
dafydd-lukes: The difference between `library()` and `require()` is stated twice in a short span of text in, I felt the phrasing should reflect this, but it's obviously a question of style preferences.

man.rmd
hadley: This is because in the printed text the link will turn into "Chapter X"

dafydd-lukes: Right, sorry, my bad!

jennybc:
After so much time, this now has merge conflicts and has been mostly implemented in PRs I just merged, working backwards. Thanks nonetheless!

PR #250, opened 2015-02-11
Fix typo in package.rmd

andrewpbray:
Appears to be a typo here.  This may or may not be the intended sentence.

"I assign the copyright of this contribution to Hadley Wickham"


hadley:
Thanks!

PR #256, opened 2015-02-16
Clarify and simplifiy 'exec' descriptor

Ironholds:
I think this is what you mean? Reads more cleanly to me.


hadley:
Sounds good!

PR #257, opened 2015-02-16
Update man.rmd

PhilipPallmann:
Fix typo.


hadley:
Wow, I'm clearly bad at counting! Thanks for catching all of these

PR #266, opened 2015-02-16
Remove one explanation of library vs require

hedderik:
Manuscript contained two almost identical explanations about the difference between library() and require(). Removed the first explanation (in the "In memory packages" section) but included a forward reference, kept the explanation in the What is a library? section.


hadley:
I like it - thanks!

PR #268, opened 2015-02-16
Update package.rmd to remove duplication.

iangow:
Discussion of difference between `library()` and `require()` was largely duplicated.


hadley:
Thanks - but I ended up going with another pull request that fixed the same problem :)

PR #270, opened 2015-02-23
backslash changed to fowardslash

robiRagan:
In the "everything comes in twos" example the slash was going the wrong way.


robiRagan:
This is hardly needed but just in case: "I assign the copyright of this contribution to Hadley Wickham"

hadley:
Thanks!

PR #280, opened 2015-03-18
Travis now supports R as a top-level language

vnijs:
I used the website for R-packages to submit my first package to CRAN. Thanks for a great resource!! I noticed that this section was out of date. My suggestions are in the proposed change. 

From the official documentation: "The R environment comes with LaTeX and pandoc preinstalled, making it easier to use packages like RMarkdown or knitr." This suggests BOOTSTRAP_LATEX=1 is no longer needed.

I didn't see any mention of "`_R_CHECK_FORCE_SUGGESTS_=0`" so I am not sure if it is still available as an option.

### DIFF
--- a/check.rmd
+++ b/check.rmd
@@ -662,78 +662,29 @@ With this setup in place, every time you push to GitHub, and every time someone 
 
 ### Basic config
 
-The Travis config is stored in a yaml file called `.travis.yml`. The default config created by devtools looks like this:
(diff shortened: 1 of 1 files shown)


jennybc:
Lots has happened since this PR, so I think it's best to just close it. The current Travis section does reflect some of the changes to the R + Travis landscape and this will also be revisited as we work on the 2nd edition. Thanks.

PR #281, opened 2015-03-30
"diamonds" -> NULL

rnuske:
According to [roxygen vignette](http://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html) and your example 'ggplot2/R/ggplot2.r' it should be NULL and not "diamonds".
I dearly missed a hint where to put that roxygen block...


jennybc:

PR #283, opened 2015-04-08
When using system.file, use mustWork = TRUE to have an error message

jomuller:
When one call internal data from the _extdata_ folder, it should be useful to have an error message when the data requested does not exist. This way, there is no need to write an error control instruction as `if (sysfile_result == "") stop("File does not exist")`

Thank you for this amazing book!

If this commit should be useful, I assign the copyright of this contribution to Hadley Wickham


jennybc:
Thanks!

PR #286, opened 2015-04-18
Update namespace.rmd

ijlyttle:
small typo


jennybc:

PR #289, opened 2015-05-01
Correct two typographical errors on Git usage

scw:
Running `git branch master` would try to create branch master (always will fail). Use `checkout` instead, Remove `extrafont` placeholder.


hadley:
Thanks!

PR #292, opened 2015-05-08
Update index.rmd

adessy:


hadley:
It reads better to me without the comma, I think.

PR #296, opened 2015-05-14
added OS X to case-insensitive operating systems

syclik:

@hadley, just adding "OS X" inside the parenthetical comment regarding case-insensitive file systems. I had that bite us while working on Stan recently. On a default configuration of the file system on a Mac, you can verify this with:
```
> touch foo
> touch Foo
> ls *oo
foo
```

(there won't be two separate files, as you'd expect on a case-sensitive file system; odd thing is that bash is case sensitive even though the file system is not)

### DIFF
--- a/r.rmd
+++ b/r.rmd
@@ -38,7 +38,7 @@ foo.r
 stuff.r
 ```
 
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #301, opened 2015-05-31
emphasize all local branches

rrunner:


rrunner:
My suggestion is not 100% correct either, since you have local branches pointing to remotes. By *all* branches, you typically use "git branch -a".

jennybc:
I think the current statement is "close enough" and this chapter will change a lot in the 2nd edition. Thanks.

PR #304, opened 2015-06-29
packageVersion("foo") > "1.0.0"

wibeasley:
instead of packageVersion("foo" > "1.0.0")

@hadley, do you prefer suggestions to be made here, or in http://www.oreilly.com/catalog/errataunconfirmed.csp?isbn=0636920034421


wibeasley:

hadley:
Thanks!

PR #305, opened 2015-06-29
accidentally being added

wibeasley:
Does *accidentally* fit better than *ever* in  "This will prevent them from accidentally being added."?  I believe you can still add files through the git command line, even if they're already in the .gitignore file.

("I assign the copyright of this contribution to Hadley Wickham".)


hadley:
Thanks!

PR #306, opened 2015-06-29
non-ASCII & encoding in Rdata files

wibeasley:
Should it be
> Data files can contain non-ASCII characters *only* if the encoding is not correctly set.

instead of
> Data files can *only* contain non-ASCII characters if the encoding is not correctly set. 

?  Fortunately I've never had to deal with this.

(I assign the copyright of this contribution to Hadley Wickham.)


hadley:
Thanks!

PR #309, opened 2015-07-06
Update tests.rmd

iargent:
Fixed missing word:
multiple cores will not always available -> multiple cores will not always be available


hadley:
Thanks!

PR #311, opened 2015-07-07
update namespace.md

jeremycg:
The anchor at "The NAMESPACE" in side links was giving an undefined destination: http://r-pkgs.had.co.nz/namespace.html#undefined

I think this was due to the given anchor {#NAMESPACE} being resolved as the same as the page title {#namespace}

I've changed the section anchor to {#the-NAMESPACE} to avoid conflict with the chapter title.


jennybc:
I picked a different anchor to disambiguate, but thanks for this catch.

PR #319, opened 2015-08-25
Fixing typos

jpinelo:
removed word "are" at end of line 68, which seems to be leftover of previous phrasing


jennybc:

PR #324, opened 2015-10-12
Update tests.rmd

perryjer1:
I think the first expect_match passes but the second fails....removing the Fail message for the first.


perryjer1:
On second thought, what I thought was a "Fail message" might have been your comment explaining why the next test would fail. If so, I found it a bit confusing.

jennybc:
OK yes it's confusing. I've added an empty line to make it less so. Thanks.

PR #327, opened 2015-11-20
`.rda` extension is case sensitive

almartin82:
Per [SO](https://stackoverflow.com/questions/21370132/r-data-formats-rdata-rda-rds-etc) and the [RHelp](http://r.789695.n4.nabble.com/rda-vs-RData-td4581445.html) list, `.RData` is the same as `.rda`.

CRAN (and `data()`) won't see files with `.Rda` extensions.  This is unfortunate, because `.Rda` appears frequently in the wild in tutorials.

### DIFF
--- a/data.rmd
+++ b/data.rmd
@@ -32,7 +32,7 @@ x <- sample(1000)
 devtools::use_data(x, mtcars)
 ```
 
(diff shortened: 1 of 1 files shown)


jennybc:
It feels like the recommendation to use `.RData` for exported data is already quite clear.

PR #334, opened 2016-01-11
Readability of expect_match example

rgknight:
I was initially confused about whether the error message was in the right place. I think this will be more clear.


hadley:
That seems more confusing to me 😕

PR #337, opened 2016-02-18
Update check.rmd

johnbaums:
Fix typo


jennybc:

PR #340, opened 2016-02-18
Update man.rmd

johnbaums:
Fix typo


jennybc:
Thanks! Fixed already via different PR.

PR #341, opened 2016-02-24
Added text about Rtools paths

gcpoole:
I was receiving error 127 when I ran has_devel().  After a lot of sleuthing, I found that the default setting when installing Rtools did not update the system path.  I have added language to the instructions for Rtools installation that warns the user to check the box to allow the installation package to update the system path.

“I assign the copyright of this contribution to Hadley Wickham


jennybc:
Thanks. I've made a note in #455 to make sure that the Rtools reality and this book are in sync by the time a 2nd edition is declared to be done.

PR #347, opened 2016-03-16
Emphasis on data file name being same as object name.

jwpestrak:

I was unable to build and check my package because the data file name was different than the object name. Maybe I'm the only person who has run into this issue, but the "same name" caveat is very important and (IMO) worth emphasis.

### DIFF
--- a/data.rmd
+++ b/data.rmd
@@ -25,7 +25,7 @@ Each possible location is described in more detail below.
 
 ## Exported data {#data-data}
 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks! I'm reluctant to do this just because the practice is not sustainable book-wide. I think many, many sentences contain someone's personal "gotcha" and we would end up with too much text in bold and italics.

PR #348, opened 2016-03-21
Update man.rmd

helix123:


jennybc:
I think the current text reflects what [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Documenting-packages) says to do and what our recommended toolchain + process produces:

> Packages may have an overview help page with an `\alias pkgname-package`, e.g. ‘utils-package’ for the utils package, when `package?pkgname` will open that help page. If a topic named pkgname does not exist in another Rd file, it is helpful to use this as an additional `\alias`.

PR #352, opened 2016-04-06
C++ development mode

krlmlr:
- More Makevars entries
- Expand unfinished paragraph

### DIFF
--- a/src.rmd
+++ b/src.rmd
@@ -205,6 +205,8 @@ To use C++ code from another package:
   `~/.R/Makevars` with:
 
     ```bash
+    LTO=-O0 # is included after all other compilation flags
+    MAKEFLAGS := -j 4 # use the number of processors available on your machine (GNU Make only)
     CC=ccache clang -Qunused-arguments
     CXX=ccache clang++ -Qunused-arguments
     CCACHE_CPP2=yes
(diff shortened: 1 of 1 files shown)


krlmlr:

jennybc:
Thanks! This has been here awhile ... I'm clearing things out in order to make a "release" to freeze the 1st edition and begin work on the 2nd. Does this need any sort of refresh?

cc @jimhester

krlmlr:
Seems that `load_all()` has improved a lot, I'm not sure the suggestion to use "Build & Reload" is still valid.

jimhester:
The described behavior of `load_all()` is not correct, it does recompile both header and object files as necessary. It can still be problematic to use `load_all()` when developing on windows due to OS locking of the shared library, but on POSIX machines it works well.

I also don't really agree with the other Makevars changes. `MAKEFLAGS` I feel is better set in your shell environment and if you want to turn off optimizations I think it is much more clear to do so in `CFLAGS`, `CXXFLAGS` etc. than using the `LTO` hack. If the `LTO` setting remains it would need much more descriptive text on why it works and why it is not an issue to override any existing LTO settings etc. Also generally I would recommend `-Og` rather than `-O0` when doing development <https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html>

jennybc:
I'm going to convert this to an issue to remind us to revisit these recommendations in the 2nd edition. Sounds like this is not a no-brainer correction type of PR that should clearly be part of the 1st edition "release".

PR #367, opened 2016-07-26
Edit explanations of package loading and attaching

dracodoc:
I was confused in reading the section `In memory packages` in chapter `Package structure`. I read the Namespaces chapter in`Advanced R` several times, then came back to read this part again. Now I found all the descriptions are accurate, but there are many names bear meanings different from what a reader may assume. I tried to edit it from the viewpoint of reader, just for your reference.

The points I tried to make are:
- Define the 3 stages `installation, loading, attaching` in very beginning. Without more detailed understanding of `loaded into memory` and `search path`, user could assume their meaning from his/her previous experience. For example he/she may think `loaded into memory` means immediately available to use, or `added to search path` happened in package installation but package is not in memory yet.
- Change the example of `devtools::install()` because any package will fit this example, while this one may add yet another similar names as cognition burden.

### DIFF
--- a/package.rmd
+++ b/package.rmd
@@ -237,7 +237,15 @@ You can prevent files in the package bundle from being included in the installed
 
 ### In memory packages
 
-To use a package, you must load it into memory. To use it without providing the package name (e.g. `install()` instead of `devtools::install()`), you need to attach it to the search path. R loads packages automatically when you use them. `library()` (and the later discussed `require()`) load, then attach an installed package.
+There are several stages involved before using a package:
+1.Package installed. 
+	With `install.packages` or `devtools::install()` etc, package is downloaded and installed, but not available to use yet.
+2.Package loaded into memory. 
+	Package can be used with prefix like `ggplot2::ggplot()`. You can explicitly load a package without attaching it with `requireNamespace()` or `loadNamespace()` but rarely need it. If you run `ggplot2::ggplot` without loading `ggplot2` package first, it will be loaded automatically. Although you still need the prefix in every usage.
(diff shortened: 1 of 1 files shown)


dracodoc:
I'm more confused about the meaning of `search path` after reading more help pages. 

- The title of `?.libPaths` is _Search Paths for Packages_, which seemed to refer package installation folder, i.e. which disk folder to lookup when package names are given.
- The help of `?library` said `library` will search through `lib.loc` (which are `.libPaths()` by default) and _load the namespace of the package with name `package` and attach it on the search list_. 

In summary, `search path` is the folder of installed packages to be looked up in loading packages. `search list` is the resulting list of the packages loaded and attached to. These two names are similar but still separatable, but they got messed up in the help page of  `?search`.

- The title of `?search` is _Give Search Path for R Objects_, previously `search path` is only for package installation folder. Using this title on help page of `search` messed up two terms.
- Furthermore, the name of `searchpaths()` and its meaning may suggest to be similar with `.libPaths()`, but it is a subset of installation folder of attached packages compare to the parent folder of all installed packages. `.libPaths` is used in the beginning of package loading, and `searchpaths()` is the result of package loading and attaching. This is a very good example of bad naming.

With my new understanding the edit I proposed should be edited further, but I'll leave them for @hadley to decide since this is more complex than my initial attempt and could have impact on more places in `R packages` and `Advance R`.

jennybc:
I'm not able to merge this multi-paragraph addition as is (unlike a typo fix). But I appreciate your point and we'll work to make this more clear in the second edition. Thanks.

PR #368, opened 2016-07-27
explanation of top-level code

dracodoc:
I didn't find a definition of `top-level code` in book or other places. From the context I'm assuming it refer to code outside functions. I tried to summary my understanding with minimal edit, but I'm not totally sure if my understanding is correct.


jennybc:
I think the meaning and proper usage of "top-level" has gotten more clear since this PR was made. Thanks!

PR #369, opened 2016-07-29
Clarify CRAN requirements for License, Authors@R, and Version

chambm:


jennybc:
I can't merge all of this PR 'as is'. If you think the book gives advice that's counter to CRAN's procedures, please open an issue and we will address it while revising for a second edition. Thanks.

PR #371, opened 2016-07-29
subtitle `The NAMESPACE` not working in content

dracodoc:
The side bar content subtitle `The NAMESPACE` is undefined. I'm not sure how it can be solved, but noticed that the {#NAMESPACE} duplicate with the first subtitle #namespace.


jennybc:
Thanks! We'll fix it a different way but I've opened an issue to track it (#453).

PR #372, opened 2016-08-05
`data.table` exception

dracodoc:
I totally agree with the recommendation of explicitly refer other package functions, but `data.table`'s special syntax inside [] may have problem if it is not imported. My edit probably didn't really explain the problem well. I just found the problem hard way, so I felt it worth to be mentioned.

I also have another problem with this general recommendation. I used lots of `stringr` methods, and using them explicitly make some very long statements, which is ugly in deep indented structure, because I have to wrap and indent a lot. I think in this case since most `stringr` method have a distinct prefix, it's very easy to notice these method came from `stringr`, thus importing `stringr` and avoiding the `stringr::` actually helps on code readability a lot.


jennybc:
We'll keep the advice here high-level and general. It's best to defer to other packages to document their own specific recommendations for how to take them on as a dependency.

PR #373, opened 2016-08-05
`Imports` assume installing CRAN version

dracodoc:
I found this in a really hard way. 

I tested my package a lot, put it to github, tried to install it to my laptop, then there were lots of errors. At first the error seemed to be `data.table` awareness problem, but further debugging point to a Shiny DT rowselected value to be character in laptop. Turned out I installed the newest github version `DT` in my PC since this is recommended by `DT` website, but my laptop only installed the CRAN version of `DT` because the Imports field only assume this.

### DIFF
--- a/description.rmd
+++ b/description.rmd
@@ -74,6 +74,8 @@ Both `Imports` and `Suggests` take a comma separated list of package names. I re
     millions of times). You'll learn about alternative ways to call functions
     in other packages in [namespace imports](#imports).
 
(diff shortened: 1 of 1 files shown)


dracodoc:
I learned later that `devtools` support specify github remote, so the problem can be solved perfectly already. However I didn't know about this from reading `R packages` book or `devtools` help pages on `use_package`, `infrastructure`.

I'd suggest to mention this feature in both `R packages` book and the help page of `use_package`.

jennybc:
We will definitely clarify this in the next edition. Thanks.

PR #383, opened 2016-10-03
Update tests.rmd

sathishsrinivasank:
added text for using helper files


sathishsrinivasank:
Hi Hadley, Your r-pkgs book is missing text on how to use helper files. I added text related to it. May be you could improve on the content or rewrite it entirely. Thanks.

PR #385, opened 2016-11-03
Update data.rmd

andrewdolman:
Documenting data sets with the roxygen format

```
#' Prices of 50,000 round cut diamonds.
#'
#' A dataset containing the prices and other attributes of almost 54,000
#' diamonds.
#'
#' @format A data frame with 53940 rows and 10 variables:
#' \describe{
#'   \item{price}{price, in US dollars}
#'   \item{carat}{weight of the diamond, in carats}
#'   ...
#' }
#' @source \url{http://www.diamondse.info/}
"diamonds"
```
causes a missing object error.

use @name tag then end with NULL


jennybc:
I think the book content reflects how data should be documented, with roxygen2 today. If you've got an example that shows otherwise, we'd welcome an issue with a reprex. Thanks.

PR #387, opened 2016-11-09
Update namespace.rmd

helix123:


jennybc:
I don't think we want to feature "Depends" prominently like this. The wording is pretty clear that `Imports` is generally where you should list the packages you depend on. We'll try to clarify this whole situation more in the 2nd edition.

PR #390, opened 2016-12-16
Update tests.rmd

kendonB:
Deleted: "You could even use an expectation within a function to check that the inputs are what you expect." This doesn't yet work nicely for testing that this in-function expectation results in an error. See here: https://github.com/hadley/testthat/issues/541


jennybc:
Agree the wording in this area could use some work, but I also don't want to just delete that sentence, w/o working on the paragraph more generally. At least the next sentence sort of makes your point:

> However, they're not so useful when something goes wrong.

We'll try to do better in the 2nd edition. Thanks.

PR #393, opened 2017-01-04
Update git.md

howbuildingsfail:
Rather than saying "RStudio preferences" (which I couldn't find), say "RStudio Tools->Global Options" - to indicate where in the RStudio menu to find the correct dialog box


jennybc:
Thanks! I changed this myself with a slightly different treatment.

Issue #349, opened 2016-03-23
use a vignette-containing example in Package structure


jennybc:
The explanations and diagram re: source vs bundled vs binary are super useful:

http://r-pkgs.had.co.nz/package.html#package

It would be even better if the example package included some vignettes, so we could track how they are represented in the different package forms.

jennybc:
Oh it's sort of there: vignettes/ --> inst/doc --> /doc. So maybe nevermind. But then why is there no build/vignette.rds in the bundle view? And the vignette source in vignettes/ persists in the bundle, but is not present in the diagram. Sorry, I have vignettes on the mind.....

jennybc:
In 2nd edition, make sure recommended vignette workflow addresses #395: the package needs to be installed in order for `browseVignettes()` to work.

hadley:
Maybe we should now include creating a vignette in the whole game chapter?

jennybc:
This appears to be covered by #587.

PR #395, opened 2017-01-10
Update vignettes.rmd

franticspider:
Added a note regarding browseVignettes to the development cycle.


jennybc:
Thanks. I think we'll give the vignette workflow a lot of thought in the 2nd edition. I've made a note of this in #349.

PR #398, opened 2017-01-29
Formatting of package names

julian-urbano:


jennybc:
Our current style does not surround package names with backticks.

PR #400, opened 2017-02-06
Specify *pkgbuild::*has_devel()

hdraisma:
Running code as in book (i.e., without this specification) caused error messages on my system:
====
> library(devtools)
> has_devel()
Error: could not find function "has_devel"
> devtools::has_devel()
Error: 'has_devel' is not an exported object from 'namespace:devtools'
> ??has_devel
> pkgbuild::has_devel()
[1] TRUE
====

### DIFF
--- a/intro.rmd
+++ b/intro.rmd
@@ -147,7 +147,7 @@ You can check that you have everything installed and working by running the foll
 
 ```{r, eval = FALSE}
 library(devtools)
-has_devel()
+pkgbuild::has_devel()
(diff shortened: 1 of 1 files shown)


jennybc:
We are in a time of upheaval in terms of the devtools diaspora. We'll make sure the book's code reflects where the dust settles before a 2nd edition goes out. I don't think we want references to `pkgbuild` in the 1st edition, which we will "freeze" soon. Thanks.

PR #335, opened 2016-01-22
Fix examples in "Exporting C code"

mllg:

PR #404, opened 2017-03-02
Fix calls to R_RegisterCCallable and R_GetCCallable

fsavje:
It seems the calls to R_RegisterCCallable and R_GetCCallable were wrong. See, e.g., https://cran.r-project.org/doc/manuals/R-exts.html#Linking-to-native-routines-in-other-packages

Updated with correct calls.


jennybc:
cc @jimhester

jennybc:
Fixed when I merged #335. Thanks for spotting!

PR #409, opened 2017-03-31
Fixed weird ub character appearing in some code chunks

simonthelwall:
In some places in the knitr section a strange \ub character had appeared. This fix removes them.


jennybc:
These are here intentionally. It's one way to present verbatim chunks for exposition.

PR #411, opened 2017-04-20
Update vignettes.rmd

davidkane9:
replace `.`` with ```

I think that this is just a typo, although perhaps it is a subtle format trick I am unaware of.


jennybc:
This is intentional. It's a trick for verbatim chunks for exposition.

PR #413, opened 2017-05-21
NEWS.md is now supported on CRAN

fsavje:


jennybc:
Thanks! I've already a merged an edit around here re: the new treatment of NEWS.md.

PR #415, opened 2017-06-10
remove extra single quotes: .onLoad, devtools.desc.author

erikerhardt:
Original code created extra quotes:
> getOption("devtools.desc.author")
[1] "\"First Last <first.last@example.com> [aut, cre]\""


jennybc:
Thanks!

PR #417, opened 2017-06-14
Remove advice to ignore NEWS.md

gregmacfarlane:
Devtools now has this warning:

    NEWS.md now supported by CRAN and doesn't need to be ignored


jennybc:
Thanks!

PR #419, opened 2017-07-25
Clarify that the check to see if you already have an SSH key-pair mus…

maurolepore:
…t be run in R.

Instruction previous to this point refer to commands that must be run in git. This point refers to a function in; but only implicitly. This may confuse users that are new to both git and R.


jennybc:
Thanks. I've made this change myself, referring to R (not RStudio).

PR #423, opened 2017-08-26
Update release.rmd

joethorley:
Added tag ` {#compatibility}` to `## Backward compatibility`


jennybc:
I can't find a place where we need this explicit label for, e.g., linking. Am I missing it?

jennybc:
Maybe you're point is that the default seems to be to label sections. I don't see any harm.

joethorley:
@jennybc yes sorry I dropped the ball on this - I wanted to link to this section but there was no tag - thanks

PR #426, opened 2017-09-23
Remove odd chars from ```{r}

talgalili:


jennybc:
This special character is an intentional device to include verbatim chunks for exposition.

PR #432, opened 2017-10-31
Update to reflect NEWS.md support by CRAN

ms609:
NEWS.md now supported by CRAN and doesn't need to be ignored.


jennybc:
Thanks! I recently merged a different change here that updates the advice.

PR #441, opened 2017-12-21
Change duplicate section anchors in Git and GitHub ch.

batpigandme:
Change anchor for "Create a local Git repository" to `{#git-local}`. Was `{git-init}`, the same as the preceding section, so the sidebar navigation takes you to "Initial set up."


jennybc:
Thanks for fixing! Can you tackle this way instead?

```
## Initial set up {#git-setup}
## Create a local Git repository {#git-init}
```

I think we'll be less likely to confuse ourselves if we keep the label `#git-init` for the section that calls `git init`

batpigandme:
Makes sense, should be all set.

jennybc:
Thanks!

PR #451, opened 2018-02-07
Update because new Rstudio

alexholcombe:
Rstudio no longer seems to have the "Rebuild and load" feature. Is the equivalent now "Clean & Rebuild"? I'm not sure.


jennybc:
Thanks. Agree it's not entirely clear which RStudio menu or button should be mentioned here. I'll make sure that the text and the IDE are in sync for a 2nd edition. It's probably safest to leave this as is for the 1st edition, which is about to get frozen.

PR #312, opened 2015-07-09
Update namespace.rmd

shntnu:
Clarified that @importFrom can also be similarly done for operators
Credits: http://stackoverflow.com/a/27979637/1094109

PR #454, opened 2018-02-09
Update namespace.rmd - importfrom can also be used for operators

shntnu:
Clarified that @importfrom can also be similarly done for operators
Credits: http://stackoverflow.com/a/27979637/1094109

Previously #312


shntnu:
@jennybc Here you go

jennybc:
Thanks!

PR #459, opened 2018-02-13
Typos

miguelmorin:
I read Hadley's second book and found two typos.


jennybc:
Thanks!

FYI for future: Although it feels silly, it is easier for us if you make one PR per change. For example in this case, the second fix has already been made by another PR. I can merge this anyway, but in general I am more likely to merge a PR if it does exactly one thing.

miguelmorin:
How naive of me: I waited until I finished the book and had all suggestions so you only had to handle one PR! Will do that in the future.

jennybc:
Yeah I didn't appreciate this until I was on the other side. But sometimes one of the changes is problematic and it's a bummer when I have to accept all or none.

PR #461, opened 2018-02-16
Corrected little typo

osorensen:
Changed, on line 250: From "However, testthat doesn't know how to cleanup after actions affect the R landscape: ..."  to "However, testthat doesn't know how to cleanup after actions that affect the R landscape: ", i.e., added "that".


jennybc:
Thanks!

PR #465, opened 2018-02-24
Corrected typo

maislind:
Changed, on line 305: From "For example is `all.equal.data.frame()` the `equal.data.frame` method for `all`, or the `data.frame` method for `all.equal`?." to "For example, is `all.equal.data.frame()` the `equal.data.frame method` for `all`, or the `data.frame` method for `all.equal`?", i.e., added a comma after sentence introduction and removed the extra period.


hadley:
Thanks!

PR #466, opened 2018-02-27
Correct missing quotation mark in Version section

malexan:


jennybc:
Hi @owenjonesuob,

Can you stop approving PRs in these repo? The 👍 is a more constructive method of expressing support and is the convention in this project. Thanks.

owenjonesuob:
Hi @jennybc, yes of course, no problem! I noticed some of these PRs had been sitting here for a long time and wondered whether I might be able to help out - if :+1: is the way to do so, then :+1: is what I'll stick to! Thanks for pointing this out.

jennybc:
This book will be next up for revision, once the new edition of Advanced R is done. So we don't want to merge much and especially don't want to take any changes re: changes in the packages related to package development.. So the repo is just lying fallow until we formally declare 1st edition frozen and reset things for 2nd edition.

owenjonesuob:
Ahhh I see, that makes perfect sense - thank you very much for the explanation! Very much looking forward to 2nd edition now 😊

hadley:
Thanks @malexan!

PR #467, opened 2018-02-27
Move sentence to following paragraph

mikemc:
Moved the sentence about using @importFrom on operators to the paragraph about using @importFrom


hadley:
Thanks!

PR #470, opened 2018-03-04
Update link to Google R style guide

cooknl:
404 error
Found this one instead
https://google.github.io/styleguide/Rguide.xml


jennybc:
Thanks. This has already been corrected in the book itself. See <http://r-pkgs.had.co.nz/r.html>. This file is sort of vestigial and will probably be pruned away in the 2nd edition prep, but it certainly can't do any harm to have a functioning URL!

PR #469, opened 2018-03-04
do *not* recommend full stop at end of title

martin-mfg:
This is to follow the tidyverse style guide: http://style.tidyverse.org/code-documentation.html#title-and-description.

I assign the copyright of this contribution to Hadley Wickham.

PR #471, opened 2018-03-05
remove full stops at ends of titles

martin-mfg:
After merging #469, the new recommendation of not putting full stops at the ends of titles should be applied to all code examples. This pull request does that.

Note that very strictly speaking, the relevant section of the tidyverse style guide only mentions functions and not datasets or other objects. But I strongly suspect that the "no full stop" rule is meant to apply to all these objects.


jennybc:
I've merged another PR that (at least partially) addresses this. And these chapters are still yet-to-be-revised, so I assume we'll catch any loose ends then. Thanks.

PR #474, opened 2018-03-22
Fix confusion of 'depends' and 'imports' 

grst:


hadley:
The change makes the text incorrect. We'll reconsider this paragraph when we come back to this chapter.

grst:
Sorry for bringing that up again, I probably should have provided more context earlier: 

Quoting from `namespace.Rmd`, [line 96](https://github.com/hadley/r-pkgs/blob/8e325a1682115b2cdf5963c4dcd34a640c13302f/namespace.Rmd#L96):
> Listing a package in either Depends or Imports ensures that it’s installed when needed. The main difference is that where Imports just loads the package, **Depends attaches it**. There are no other differences.

Quoting from `namespace.Rmd`, [line 269](https://github.com/hadley/r-pkgs/blob/8e325a1682115b2cdf5963c4dcd34a640c13302f/namespace.Rmd#L269):
> Depends is just a convenience for the user: if your package is attached, it also attaches all packages listed in Depends. If your package is loaded, packages in **Depends are loaded, but not attached**, so you need to qualify function names with :: or specifically import them.

This is obviously a contradiction. 
From what I can tell from other sources [[1]](https://stackoverflow.com/questions/8637993/better-explanation-of-when-to-use-imports-depends), [[2]](https://cran.r-project.org/doc/manuals/R-exts.html#Package-Dependencies), the first version is correct.

PR #475, opened 2018-03-29
Change has_devel() to has_compiler() in intro.rmd

kokbent:
Couldn't find any has_devel() following the instructions given. Instead I found has_compiler() under `pkgbuild`


jennybc:
Yes, we have a lot of modernization to do in a 2nd edition in terms of the current state of all the relevant packages.

So I won't handle something like this until we've officially put 1st edition to bed (made a release / tag), then reset for next edition, FYI.

hadley:

PR #476, opened 2018-04-04
Updated requireNamespace()

osorensen:
`requireNamespace(x, quitely = TRUE)` will not work, since the first argument must be a string. Added quotes around `x`.


hadley:
Thanks!

PR #478, opened 2018-04-09
Move `context()` call after `library()` call

jimhester:
Having the `library()` call first gives you an error in the CRAN version of testthat.

Fixes https://github.com/r-lib/testthat/issues/700#issuecomment-379578564


jennybc:
Thanks!

PR #493, opened 2018-07-10
Minor change

georoen:
Matching version number of example code with surounding text ("For example, imagine your package is version `0.5.0` and you want to remove `fun()`. In version, `0.6.0`, you'd use `.Deprecated()` to display a warning message whenever someone uses the function").


jennybc:
Thanks!

PR #505, opened 2018-10-23
devtools::use_data is deprecated.

jomuller:
`usethis::use_data` instead


hadley:

PR #510, opened 2018-11-14
Update release.rmd

jjesusfilho:
replaced `devtools::build_win()` by `devtools::check_win_*()` at line 218.


hadley:

PR #511, opened 2018-11-15
system.file add arg mustWork +  devtools::use_vignette is deprecated

jomuller:
- When using system.file, use mustWork = TRUE to have an error message 
- devtools::use_vignette is deprecadted


hadley:

PR #512, opened 2018-12-06
devtools::use_testthat() is deprecated in tests

valeonte:
Replaced the usage with usethis::use_testthat() as recommended from the deprecated message


hadley:

PR #513, opened 2018-12-06
Typos and updating to latest output in Vignettes

valeonte:
devtools::use_vignette is deprecated, so replaced with the recommended one. Removed some magic characters occasionally appearing next to `s. Updated template vignette generated to reflect to what is actually is generated when using latest version of packages


jennybc:
Please see the note in [README](https://github.com/hadley/r-pkgs#readme) re: why this will just sit here.

hadley:

PR #515, opened 2018-12-19
Clarifying the .Rbuildignore example contents

romanzenka:
I copy&pasted your sample code only to see R ignoring my changes completely. I eventually figured out the problem is in those comments, so I updated your documentation to be less misleading.


jennybc:
Thanks! I just did a major revision to that chapter (126038c6e45f6d81c29ec5a705d56d9e2b155203) and one of the things I clarified is this issue. I won't be able to merge this because your base and `master` have now diverged considerably.

PR #516, opened 2018-12-22
Update vignettes.rmd - replace devtools::use_vignette() with usethis::use_vignette

simonthelwall:
`devctools::use_vignette()` is deprecated, so have edited this to `usethis::use_vignette()`


jennybc:
Please see the README for this repo re: why PRs like this are likely to not be acted upon in the near future.

hadley:

PR #517, opened 2019-01-12
Replace deprecated devtools call

MichaelChirico:
`devtools::use_vignette` has a deprecation warning and then calls `usethis::use_vignette`; reflecting that here


jennybc:
Please see the README for some context re: why we aren't going to tackle this problem at this moment and / or in a piecemeal way.

MichaelChirico:
@jennybc thanks, filed this quickly and missed the `README`. Shall I close this? I take it the plan is to do a huge overhaul at some point (near-mid future I guess)...

jennybc:
It's fine. Either these issues will be helpful (there are many others) or we'll close *en masse* when the work starts.

PR #522, opened 2019-01-29
Matching parentheses

wibeasley:
I'm assuming the starting parenthesis was missing, instead of the closing parenthesis was duplicated (from the hyperlink).


hadley:
Thanks, but we're going to delete this chapter in favour of http://happygitwithr.com/

PR #524, opened 2019-02-02
Fixed Typo

mayankvanani:
removed additional(dot) character in line 237 , 269 and 275


hadley:
That's placed there deliberately to avoid rendering the chunks.

mayankvanani:
Yeah, I figured when the travis built failed.
Thanks.

PR #535, opened 2019-02-12
Use Pandoc div syntax for (aspirational) tips about RStudio and workf…

jennybc:
…low/lifestyle

tagging you in @hadley


hadley:
Note that I don't see the images in the RStudio preview, but I do when I pop the chapter out in to the browser.

style.css
```diff
@@ -12,3 +12,45 @@ pre {
...
+  background-position: 5px center;
+}
+
+div.rstudio-tip {
+  background-image: url("../diagrams/lightbulb.png");
+}
+
+div.tip {
+  background-image: url("../rstudio.png");
+
```
jennybc: I assume this empty line can go?

style.css
```diff
@@ -12,3 +12,45 @@ pre {
...
+}
+
+div.base::before {
+  display: block;
+  content: "In base R";
+}
```
jennybc: Do you think we'll use these too or should we delete this?

hadley:
I think this is good to go — we can certainly tweak the images and add more options in the future.

PR #536, opened 2019-02-18
make clear you can start from any open R console

Br-Johnson:
ie. no need to start a new project first or anything (so used to having to do that)


jennybc:
Thanks!

PR #542, opened 2019-02-26
Update intro.Rmd

sibusiso16:
fixed what i thought was not correct


jennybc:
Thanks, but we are doing our own overhaul on the intro.

PR #543, opened 2019-03-07
devtools::use_testthat() -> usethis::use_testthat()

simdadim:
Changed to usethis::use_testthat() as is the new recomendation


jennybc:
Thanks for pointing out -- there is a lot of this sort of work to be done, with all the changes in devtools, and it's possible that these sorts of PRs (there are many) will get subsumed by some bulk action that I do.

jennybc:
Thanks but I've just addressed a lot of this *en masse*.

PR #544, opened 2019-03-08
Minor fix, replace deprecated `build_win()` refs

brews:


jennybc:
Thanks, but I've just taken care of this as part of a larger effort.

PR #547, opened 2019-03-14
Replace mention of .Rd files with NAMESPACE file

JayCeBB:
Replace mention of .Rd files with NAMESPACE file in Roxygen workflow.


jennybc:
Well, the generation of the `.Rd` files is awfully important here. Instead of complete replacement, I think mention both.

PR #549, opened 2019-04-03
write fcount_fodder to fcount.R (not fcount_fodder.R)

apreshill:
Hi @jennybc,

I *think* this is a typo in the Whole Game chapter, in the [section](https://r-pkgs.org/whole-game.html#use_package) where you demo `use_package`.

This is the reason `fcount.R` is empty here: 
https://github.com/jennybc/foofactors/blob/master/R/fcount.R

But you do have `fcount_fodder.R` here: 
https://github.com/jennybc/foofactors/blob/master/R/fcount_fodder.R

See this [commit](https://github.com/jennybc/foofactors/commit/83ee478061e5234805f76b1e2c488e79a2628588).

Alison

### DIFF
--- a/Whole-game.Rmd
+++ b/Whole-game.Rmd
@@ -614,7 +614,7 @@ cat(fcount_fodder, sep = "\n")
 ```
 
 ```{r write-fcount, include = FALSE, eval = create}
-writeLines(fcount_fodder, path("R", "fcount_fodder.R"))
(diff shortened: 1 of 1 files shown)


jennybc:
Yeah, I think you're right. I'll merge this and then eventually I'll remake the GitHub example. Thanks!

PR #558, opened 2019-05-23
devtools::use_testthat() is deprecated.

calligross:


calligross:
Someone else already made an identical pull request

PR #559, opened 2019-05-24
Update all links eligible to HTTPS

Bisaloo:
I manually checked all links before changing them to ensure HTTPS was fully functional and did not cause any mixed content blocking issues.

FYI, I also noticed several dead links and links to the older version of this book in the process.


jennybc:
Thanks! It was a hairy merge, because many things have changed since this PR was initiated, but I believe I straightened it out.

PR #568, opened 2019-06-30
Fixing the last two instances of devtools:: -> usethis::

gustavdelius:


jennybc:
Thanks! I commented on your devtools PR btw

 I think these should possibly not be namespace ... but if they are, it should be `usethis::`, so will merge.

PR #571, opened 2019-07-04
structure.Rmd: Image files not shown in the book

statnmap:
The following code does not allow to render images in the book
```
<img width = "30%" src = "images/project-options-1.png"/><img width = "70%" src = "images/project-options-2.png"/>
```
I suggest using `knitr::include_graphics()` as in the rest of the book.
If the two images need to be side by side, it would be preferable to create a unique image.


jennybc:
Thanks!

PR #573, opened 2019-07-13
Deleted a couple of lines from vignette metadata

chsafouane:
The default vignette (the one generated by usethis::use_vignette) doesn't include "author" and "date" info in YAML section


jennybc:
Thanks!

PR #575, opened 2019-07-21
Add section on checking if package name is available using the availa…

### LINKED ISSUE #574: mention `available::available` in the book?
The [`available` package](https://github.com/ropenscilabs/available) is awesome for checking if your R package is available on CRAN, bioconductor, and github. It'd be great to mention it in the R packages book, and I'm happy to submit a PR for this that says something like the following:

```
It is a good idea to spend some time choosing your package name.
You can check if your package name is available for use on CRAN 
by using the [`available` package](https://github.com
ropenscilabs/available), which does the following:

* Checks for validity
* Checks not already available on GitHub, CRAN and Bioconductor
* Searches Urban Dictionary, Wiktionary and Wikipedia for 
unintended meanings

(insert screenshot here?)

While checking if your package name is available, it might also 
be worthwhile to check if the functionality of your package 
already exists - you may save yourself a lot of time using 
existing software rather than writing your own.
```

But I'm not sure where to put it - perhaps:

* at the [start or end end of whole game](https://r-pkgs.org/whole-game.html#the-end-check-and-install)
* in the [section on package naming](https://r-pkgs.org/package-structure.html#naming)
* or in the [release section](https://r-pkgs.org/release.html)

Do you have any thoughts on this?

Also see: https://github.com/r-lib/usethis/issues/808

njtierney:
…ble package.

* Also adds a section from https://devguide.ropensci.org/building.html#naming-your-package with a link to two blog posts I wrote about this (feel free to remove!)
* Closes #574


jennybc:
Thanks! I had to monkey with your `master` branch to make this mergeable 😱 so you have been warned. If I were you, I'd probably nuke my fork and re-fork next time you make a PR and ... *inserting a saved reply*

FYI in the future, we'd prefer to get a pull request from a non-`master` branch. It's generally acknowledged as a better practice to make PRs from a feature branch and it has workflow benefits for both parties (PR maker and receiver).

njtierney:
Thanks, Jenny! So sorry for creating more work for you 🙀  I will do my best in the future to go from a non-master branch!

PR #581, opened 2019-08-08
Clarify '16.6 Release' section

SangdonLim:
* Clarify how `cran-comments.md` is processed in **16.6 Release**.

I was confused about whether `cran-comments.md` is supposed to be included in the uploaded package bundle. I hope this clarifies it.

### DIFF
--- a/release.Rmd
+++ b/release.Rmd
@@ -423,8 +423,10 @@ You're now ready to submit your package to CRAN. The easiest way to do this is t
     can't otherwise be automated.
   
 * Uploads the package bundle to the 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks, but I feel like this is already covered when the idea of `cran-comments.md` is introduced:

https://github.com/hadley/r-pkgs/blob/5af3a13cc2eae4af1bd8ca47d417833d55ebf6d0/release.Rmd#L157

The `release()` function has fallen somewhat out of favour (although I'm still a big fan). So I'm going to close this because I think current state is OK and this section has to be heavily revised anyway.

PR #583, opened 2019-08-21
remove suggestion about `as.person(...)`

pearsonca:
Agree that `as.person` exists and works nicely, but per a rejected package update, CRAN now seems to require `person(...)` or `c(person(...), person(...), ...)`.  Using `as.person(...)` generates a note on their automated system:

```
* checking CRAN incoming feasibility ... NOTE
Authors@R field should be a call to person(), or combine such calls.
```

### DIFF
--- a/metadata.Rmd
+++ b/metadata.Rmd
@@ -261,17 +261,6 @@ Authors@R: c(
     person("Winston", "Chang", email = "winston@rstudio.com", role = "aut"))
 ```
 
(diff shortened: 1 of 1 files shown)


pearsonca:
I can confirm that just changing `as.person(c(...))` got my submission through.

Not sure about the CI check failing notice, but looks like a pipeline config issue, not to do with proposed deletion.

jennybc:
Thanks, looks good.

PR #596, opened 2019-09-11
R_RegisterCCallable: package name instead of DLL pointer

agrueneberg:
`R_RegisterCCallable` takes the package name as its first argument, not the `DllInfo` pointer.


jennybc:
The content on compiled code is slated to become more minimal and link to external resources, since this is a topic that really demands its own, thorough treatment. So I'm closing PRs that act on the existing text, but thanks.

PR #598, opened 2019-09-16
Replace all instances of vertical tab with unicode zero width spaces

### LINKED ISSUE #597: Code block typo
https://r-pkgs.org/vignettes.html#knitr

\`\`\`{r}

should be 

 \`\`\`{r}

jimhester:
The vertical tabs show up in the HTML as whitespace, the zero width
spaces do not

Fixes #597


hadley:
Thanks — I'm surprised they weren't already ZWS as that's what I remember using

PR #610, opened 2019-11-13
s/devtools/revdepcheck/

chezou:
Since `devtools` doesnt have revdep_check() anymore, we need to use revdepcheck instead.


jennybc:
Thanks! This content has been overhauled since this PR came in and is effectively addressed in the current version.

PR #614, opened 2019-12-20
Installing with vignettes should have "dependencies=TRUE"

Marlin-Na:
Refer https://github.com/r-lib/devtools/issues/2168


hadley:
Lets fix this in remotes instead: https://github.com/r-lib/remotes/issues/709

PR #615, opened 2020-01-10
minor grammar tweak

MichaelChirico:
Currently it reads like `NAMESPACE` is a directory. Alternative:

> the `NAMESPACE` file and files in the `man/` directory

### DIFF
--- a/git.Rmd
+++ b/git.Rmd
@@ -306,7 +306,7 @@ knitr::include_graphics("images/git-ignore.png", dpi = 220)
 
 If you want to ignore multiple files, you can use a wildcard "glob" like `*.png`. To learn more about the options, see [ignoring files](http://git-scm.com/book/ch2-2.html#Ignoring-Files) in Pro-Git.
 
-Some developers never commit derived files, files that can be generated automatically. For an R package this would mean ignoring the files `NAMESPACE` and anything in the `man/` directory because they're generated from comments. From a practical pespective, it's better to commit these files: R packages have no way to generate `.Rd` files on installation so ignoring derived files means that users who install your package from GitHub will have no documentation.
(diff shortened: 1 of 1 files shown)


git.Rmd
```diff
@@ -306,7 +306,7 @@ knitr::include_graphics("images/git-ignore.png", dpi = 220)
 
 If you want to ignore multiple files, you can use a wildcard "glob" like `*.png`. To learn more about the options, see [ignoring files](http://git-scm.com/book/ch2-2.html#Ignoring-Files) in Pro-Git.
 
-Some developers never commit derived files, files that can be generated automatically. For an R package this would mean ignoring the files in the `NAMESPACE` and `man/` directories because they're generated from comments. From a practical pespective, it's better to commit these files: R packages have no way to generate `.Rd` files on installation so ignoring derived files means that users who install your package from GitHub will have no documentation.
+Some developers never commit derived files, files that can be generated automatically. For an R package this would mean ignoring the files `NAMESPACE` and anything in the `man/` directory because they're generated from comments. From a practical pespective, it's better to commit these files: R packages have no way to generate `.Rd` files on installation so ignoring derived files means that users who install your package from GitHub will have no documentation.
```
jimhester: Maybe this?

```suggestion
Some developers never commit derived files, files that can be generated automatically. For an R package this would mean ignoring the `NAMESPACE` file and anything in the `man/` directory because they're generated from comments. From a practical perspective, it's better to commit these files: R packages have no way to generate `.Rd` files on installation so ignoring derived files means that users who install your package from GitHub will have no documentation.
```

MichaelChirico: Works for me, thanks

MichaelChirico:
A few more observations on looking at this again:

 1. It might be good to mention the word `artifact`, which I've seen used more often than `derived files`
 2. I have more than once recently gotten pushback that "`man` shouldn't be tracked on GitHub" for the reason given here. Given that `devtools` and `roxygen2` themselves are tracking `man`, it might be nice to have a more full-throated defense of the practice here to be able to refer to (currently, it come across as slightly stronger than indifferent)

jennybc:
Thanks, but the Git content here is going to be scaled back and link to external resources. So it doesn't make sense to groom it at this point. This will all go away before too much longer.

PR #616, opened 2020-01-19
Deleted context() recommendation

chsafouane:
I believe it's not longer recommended to use `context()`. It is mentioned elsewhere in the chapter but I believe It'd better be adapted by the author

### DIFF
--- a/testing.Rmd
+++ b/testing.Rmd
@@ -393,7 +393,7 @@ test_that("floor_date works for different units", {
 
 ## Test files {#test-files}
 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks. I might not merge this because, as you say, I prefer to update a larger unit at once. I know this needs updating but it's easier (and probably less confusing to readers?) to update a whole chapter or section in one go.

chsafouane:
Thanks Jenny. I think you are 100% right. As you are aware of the need to update the chapter, I'm going to close the pull request

PR #627, opened 2020-04-04
added info that markdown is supported

tjebo:
I am happy to add some more examples if this is desired. 


jennybc:
Thanks, but this chapter hasn't been revised yet. There will be considerable change re: the current roxygen2 landscape, so I'm closing this because it feels better to update the whole chapter at once versus piecemeal.

PR #628, opened 2020-04-04
Fix link in Introduction.Rmd

Lenostatos:
It seems to me that the link in the third paragraph of section 1.2 at https://r-pkgs.org/intro.html should not point to chapter 17.2.2 but to chapter 4. I'm not very experience with RMarkdown so I don't know whether I changed the document correctly. I also don't know how to build this book in order to test it.


jennybc:
Thanks! Looks right to me.

PR #634, opened 2020-05-08
Remove full stop in examples' title

espinielli:
Following the "suggestion" as in:

> * The first _sentence_ becomes the title of the documentation. That's what you 
>  see when you look at `help(package = mypackage)` and is shown at the top of 
>  each help file. It should fit on one line, be written in sentence case, but not 
>  end in a full stop.

I removed the full stop from the examples in this chapter.

Not sure about S4 and RC so I haven't touched those.


jennybc:
Thanks!

PR #641, opened 2020-07-18
Fix URL for R-devel for Mac

brianrice2:
Moved from the old att.com location (though it still redirects okay). "This is the new home for experimental binaries and documentation related to R for macOS." Neighboring links look okay.

Copyright requirements sound good to me&mdash;I assign the copyright of this contribution to Hadley Wickham.


jennybc:
Thanks!

PR #648, opened 2020-08-18
Revise Code.Rmd (deals with `R/`)

jennybc:


Code.Rmd
```diff
@@ -7,19 +7,30 @@ source("common.R")
...
 
-The first practical advantage to using a package is that it's easy to re-load your code.
-You can either run `devtools::load_all()`, or in RStudio press __Ctrl/Cmd + Shift + L__,  which also saves all open files, saving you a keystroke.
+As your use of R becomes more sophisticated, it's common to start to write your own R functions.
+If a function is only used in one place, you probably define it right there.
+But if you've bothered to write a function, it's likely you want to reuse it in multiple places: within one script, across multiple scripts, or even across multiple projects.
+
+The next logical step is to collect such function definitions in one or more dedicated `.R` files and then `source()` them as needed.
+Typically these functions co-evolve with the code where you use them, i.e. your analysis code, your Shiny apps, or your R Markdown reports.
+
```
jennybc, line 21: I felt like this needed more of a intro to set the stage.

Something that felt very *implicit* before is that the user has "their code" and that cleaves into 2 types of code: top-level code that carries out their analysis (or whatever) and functions they have defined. And it's the *latter* that we're proposing to store in a package.

Code.Rmd
```diff
@@ -7,19 +7,30 @@ source("common.R")
...
+:::rstudio-tip
+RStudio exposes `load_all()` in the *Build* menu, in the *Build* pane via *More > Load All*, and in keyboard shortcuts Ctrl + Shift + L (Windows & Linux) or Cmd + Shift + L (macOS). When run inside RStudio `load_all()` also saves all open files, saving you a keystroke.
+:::
 
 This keyboard shortcut leads to a fluid development workflow:
 
 1. Edit an R file.
 
-1. Press Ctrl/Cmd + Shift + L.
+1. Execute `load_all()`, probably via Ctrl/Cmd + Shift + L.
```
jennybc, line 33: Make it *slightly* less RStudio-specific.

Code.Rmd
```diff
@@ -7,19 +7,32 @@ source("common.R")
...
+You'll also learn some important differences between defining a function in a script and defining it in a package.
 
 ## R code workflow {#r-workflow}
 
-The first practical advantage to using a package is that it's easy to re-load your code.
-You can either run `devtools::load_all()`, or in RStudio press __Ctrl/Cmd + Shift + L__,  which also saves all open files, saving you a keystroke.
+As your use of R becomes more sophisticated, it's common to start to write your own R functions.
+If a function is only used in one place, you probably define it right there.
+But if you've bothered to write a function, it's likely you want to reuse it in multiple places: within one script, across multiple scripts, or even across multiple projects.
+This is *exactly* what an R package is for!
```
jennybc, line 18: I felt like this needed more of a intro to set the stage.

Something that felt very *implicit* before is that the user has "their code" and that cleaves into 2 types of code: top-level code that carries out their analysis (or whatever) and functions they have defined. And it's the *latter* that we're proposing to store in a package.

Code.Rmd
```diff
@@ -7,19 +7,32 @@ source("common.R")
...
+
+Without package technology, you probably collect these function definitions in one or more dedicated `.R` files and then `source()` them as needed.
+Typically these functions co-evolve with the code where you use them, i.e. your analysis code, your Shiny apps, or your R Markdown reports.
+If you use these functions across multiple projects, you also face the uncomfortable dilemma of where to define them and whether to have multiple, slightly different copies of this code lying around.
```
jennybc, line 22: This feels like a second point, that is:

* worth making
* possibly doesn't belong *right here* (but it sort of does)

Code.Rmd
hadley: Yeah, I like it!

Code.Rmd
hadley: Although it begs the question of how you manage your local package installs, versioning etc.

Code.Rmd
hadley: I'd rather keep the keyboard shortcut primary, because it is such an important part of the workflow.

Code.Rmd
```diff
@@ -29,23 +42,126 @@ Congratulations!
 You've learned your first package development workflow.
 Even if you learn nothing else from this book, you'll have gained a useful workflow for editing and reloading R code.
 
-## Organising your functions
+## Organising your functions into files {#r-organising}
+
+*We originally decided to remove this in deference to material in <https://style.tidyverse.org>. The removed text is in [tidyverse/style/#121](https://github.com/tidyverse/style/issues/121). In hindsight, I think removal is too extreme and I've re-introduced this section. First, the original material really isn't in style.tidyverse.org (yet?). Second, people really want guidance on this and it feels misguided to omit from this book out of some adherence to the DRY principle.*
```
hadley, line 47: Agreed

Code.Rmd
```diff
@@ -29,23 +42,126 @@ Congratulations!
...
+
+The only hard rule is that your package should store its function definitions in R scripts, i.e. files with extension `.R`, that live in the `R/` directory[^no-subdirs].
```
hadley, line 49: I think it's worth making the distinction that R doesn't care, but human readers will.

Code.Rmd
```diff
@@ -29,23 +42,126 @@ Congratulations!
...
+This advice should inform your general policy, but there are exceptions to every rule.
+If a specific function is very large or has lots of documentation, it can make sense to give it its own file, named after the function.
+More often, a single `.R` file will contain multiple function definitions: such as a main function and its supporting helpers, a family of related functions, or some combination of the two.
+
+Here are some examples from the actual source of the [tidyr package](http://tidyr.tidyverse.org/):
+
+```{r include = FALSE}
+library(tidyverse)
+
+# TODO: should these links be SHA'd to guard against change?
```
hadley, line 69: Yes, especially since they make me see that I don't like the current organisation

Code.Rmd
```diff
@@ -29,23 +42,126 @@ Congratulations!
...
+```{r echo = FALSE}
+knitr::kable(dat)
+```
+
+:::tip
+Another file you often see in the wild is `R/utils.R`.
+This is a common place to define small utilities that are used inside multiple package functions.
+Since they serve as helpers to multiple functions, placing them in `R/utils.R` makes them easier to re-discover when you return to your package after a long break.
+
+*TODO: link to Bob Rudis's blog post? <https://rud.is/b/2018/04/08/dissecting-r-package-utility-belts/>*
```
hadley, line 101: :+1:

Code.Rmd
```diff
@@ -29,23 +42,126 @@ Congratulations!
...
+  the list and eventually pick a function (or file) to visit. This works for both
+  functions and files in your project.
+* With your cursor in a function name or with a function name selected, press
+  F2. This works for functions defined in your package or in another package.
+  
+After navigating to a function with one of these methods, return to where you started by clicking the back arrow at the top-left of the editor or by pressing Ctrl + F9 (Windows & Linux) or Cmd + F9 (macOS).
+:::
 
-*removed in deference to material in <https://style.tidyverse.org>; see [tidyverse/style/#121](https://github.com/tidyverse/style/issues/121)*
+*TODO: I've let two images go here because unclear how that would/should work inside an RStudio tip. Are we OK with this? See them in the 1st ed at the end of this section: <http://r-pkgs.had.co.nz/r.html#r-organising>*
```
hadley, line 118: I think you should keep them. Why would they work differently inside a tip?

Code.Rmd
```diff
@@ -29,23 +42,126 @@ Congratulations!
...
+* `styler::style_file()` restyles a single file.
+* `styler::style_text()` restyles a character vector.
+
+:::rstudio-tip
+When styler is installed, the RStudio Addins menu will offer several additional ways to style code:
+
+* the active selection
+* the active file
+* the active package
+:::
```
hadley, line 142: Mention rebinding RStudio's default re-style keyboard shortcut? (that's what I do)

Code.Rmd
```diff
@@ -131,37 +259,148 @@ There are some functions that modify global settings that you should never use b
...
-  
-    ```{r, eval = FALSE}
-    old <- options(stringsAsFactors = FALSE)
-    on.exit(options(old), add = TRUE)
-    ```
+Sometimes you cannot avoid modifying the state of the world, in which case you just have to make sure that you put things back the way you found them.
 
-* Avoid modifying the working directory. If you do have to change it, make sure
-  to change it back when you're done:
+The first function to know about is base R's [`on.exit()`](https://rdrr.io/r/base/on.exit.html).
```
hadley, line 266: I think it would be better to start with the specific withr functions, then talk about `withr::defer()`, and then link to something that explains `on.exit()`. The only downside is that might force a discussion about dependencies earlier than you might otherwise want it.

Code.Rmd
```diff
@@ -173,25 +412,27 @@ For now, you should always use `.onLoad()` unless explicitly directed otherwise.
 
 Some common uses of `.onLoad()` and `.onAttach()` are:
 
-*   To display an informative message when the package loads. This might make 
-    usage conditions clear, or display useful tips. Startup messages is one 
-    place where you should use `.onAttach()` instead of `.onLoad()`. To display 
-    startup messages, always use `packageStartupMessage()`, and not `message()`. 
-    (This allows `suppressPackageStartupMessages()` to selectively suppress 
-    package startup messages).
+* To display an informative message when the package loads. This might make
```
hadley, line 415: BTW I prefer greater indenting here so that the code and bullets line up nicely.

Code.Rmd
```diff
@@ -173,25 +412,27 @@ For now, you should always use `.onLoad()` unless explicitly directed otherwise.
...
     .onAttach <- function(libname, pkgname) {
       packageStartupMessage("Welcome to my package")
     }
     ```
     
-*   To set custom options for your package with `options()`. To avoid conflicts
-    with other packages, ensure that you prefix option names with the name
-    of your package. Also be careful not to override options that the user
-    has already set.
+* To set custom options for your package with `options()`. To avoid conflicts
```
hadley, line 428: Do we still believe in this? Or should you define helpers to get options that you can document, and that provide defaults?

Code.Rmd
```diff
@@ -212,17 +453,16 @@ Some common uses of `.onLoad()` and `.onAttach()` are:
...
-    make C++ classes available as reference classes in R with Rcpp modules,
-    you call `Rcpp::loadRcppModules()`. 
-
-*   To register vignette engines with `tools::vignetteEngine()`.
+* To connect R to another programming language. For example, if you use rJava to
+  talk to a `.jar` file, you need to call `rJava::.jpackage()`. To make C++
+  classes available as reference classes in R with Rcpp modules, you call
+  `Rcpp::loadRcppModules()`. 
 
+* To register vignette engines with `tools::vignetteEngine()`.
```
hadley, line 465: I think we can probably omit this example

hadley: Maybe replace with a forward link handling different versions of R/dependencies?

Code.Rmd
jennybc: I feel like that's what the very next sentence does 👇 It's certainly the intent

jennybc:
Closing in favour of #659. The history got very messy here.

PR #661, opened 2020-09-15
typo fix

maelle:
_Very simple contributions like typo fixes are generally not protected by copyright because they’re not creative works. _ :wink:


jennybc:
Thanks!

PR #669, opened 2020-10-06
Standardise workflow

hadley:


.github/workflows/build-book.yaml
```diff
@@ -23,29 +23,15 @@ jobs:
 
       - name: Setup R
         uses: r-lib/actions/setup-r@master
-        with:
```
hadley, line 26: I am almost certain this didn't help the original problem, since it only sets the option for that one action

.github/workflows/build-book.yaml
```diff
@@ -23,29 +23,15 @@ jobs:
...
-          crayon.enabled: 'FALSE'
-          
-      - uses: r-lib/actions/setup-pandoc@master
 
-      - name: Get R version
```
hadley, line 31: https://github.com/r-lib/actions/commit/79b0ee591d1cdce07ab0ff6a2519af150bd4390b

.github/workflows/build-book.yaml
```diff
@@ -23,29 +23,15 @@ jobs:
...
+      - name: Setup pandoc
+        uses: r-lib/actions/setup-pandoc@master
 
       - name: Cache R packages
         uses: actions/cache@v2
         with:
           path: ${{ env.R_LIBS_USER }}
-          key: ${{ runner.os }}-${{ steps.get-r-version.outputs.version }}-1-
-
-      - name: Cache bookdown results
```
hadley, line 43: Moved to the bottom so it's next to the action that it applies to

.github/workflows/build-book.yaml
jennybc: Screen Shot 2020-10-06 at 3 20 32 PM

Nope. Screenshot of the successful deployment after I bumped the bookdown cache.

hadley: But how does it work? Did @jimhester add something specific to the setup-R action?

jennybc: Yes, that is my impression:

https://github.com/r-lib/actions/blob/79b0ee591d1cdce07ab0ff6a2519af150bd4390b/setup-r/action.yml#L14

And that makes its way into options set via an `.Rprofile`:

https://github.com/r-lib/actions/blob/79b0ee591d1cdce07ab0ff6a2519af150bd4390b/setup-r/src/installer.ts#L365

https://github.com/r-lib/actions/blob/79b0ee591d1cdce07ab0ff6a2519af150bd4390b/setup-r/lib/installer.js#L359

I added this originally based on research on the action and what others are doing in similar bookdown projects.

hadley: Ah, it's a lack of documentation. I've filed an issue.

hadley: Thinking about this more, I do think the right place for this setting is close to the book generation, since the root cause is that bookdown doesn't handle ANSI escapes. (But that's something that I will hopefully fix in the next month, rendering the point moot)

.github/workflows/build-book.yaml
```diff
@@ -76,6 +64,12 @@ jobs:
           git config --local user.name "$GITHUB_ACTOR"
           git config --local user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
+      - name: Cache bookdown results
+        uses: actions/cache@v2
+        with:
+          path: _bookdown_files
+          key: bookdown-8-${{ hashFiles('**/*Rmd') }}
```
hadley, line 71: I removed the restore key here — I think that may have been exacerbating the caching issue.

PR #670, opened 2020-10-06
Put all env vars in same place

hadley:
Opening for the purposes of discussion


jennybc:
Seems fine to me. The CI failure is the always-infuriating cache problem.

PR #674, opened 2020-10-29
Use bs4_book template

hadley:


DESCRIPTION
```diff
@@ -21,12 +22,20 @@ Imports:
     usethis,
     withr
 Suggests:
+    bookdown,
```
hadley, line 25: FWIW the distinction I use for imports vs suggests for books is that imports is need to run the code in the book; suggests is needed to build the book itself

_output.yml
```diff
@@ -1,3 +1,12 @@
+bookdown::bs4_book:
+  theme:
+    primary: "#982a31"
```
hadley, line 3: I don't know if we need to formally record this anywhere but these colours were produced by taking O'Reilly red (#BF353F) and setting chroma to 80 and lightning to 35 for the primary colour, and chroma to 5 and lightness to 14 for the text colour.

DESCRIPTION
```diff
@@ -22,19 +24,18 @@ Imports:
...
+    downlit,
+    jsonlite,
     parsermd,
-    pkgload
-Remotes:  
+    pkgload,
+    xml2
+Remotes: 
     r-lib/devtools,
     r-lib/gert,
```
jennybc, line 36: We are probably fine with CRAN gert now

hadley:
[Whole game](https://5f9f3ad97330086a15debcea--r-pkgs.netlify.app/whole-game.html) looks good to me (although I don't see any colour), so I think this is good to merge. As long as any problems are spotted in the next ~2 weeks I should be able to fix quickly.

jennybc:
I see the ANSI escapes:

Screen Shot 2020-11-02 at 6 56 37 AM

hadley:
Ok, I see it too at https://5f9f504f2fba21013795d11a--r-pkgs.netlify.app/whole-game.html#create_package

PR #675, opened 2020-10-29
Tweak caching for whole game special needs

hadley:


Whole-game.Rmd
```diff
@@ -757,9 +768,8 @@ cat(pretty_install(), sep = "\n")
 Feel free to visit the [foofactors package](https://github.com/jennybc/foofactors) on GitHub, which is exactly as developed here. The commit history reflects each individual step, so use the diffs to see the addition and modification of files, as the package evolved. The rest of this book goes in greater detail for each step you've seen here and much more.
 
 ```{r cleanup, include = debug, eval = create}
-if (requireNamespace("foofactors", quietly = TRUE)) {
-  remove.packages("foofactors")
-}
+pkgload::unload("foofactors")
```
jennybc, line 771: We should add pkgload to DESCRIPTION.

Package-within-files/echo-cleaning-helpers.R
```diff
@@ -1,4 +1,4 @@
-lookup_table <- tribble(
+lookup_table <- dplyr::tribble(
```
jennybc, line 1: Yes, this is correct / makes sense to me, based on a small fix made recently.

PR #678, opened 2020-11-23
Improve package cache

hadley:
Now always updates to latest version of CRAN packages, and caches all versions of dependencies


jennybc (approved):
I have every confidence that this is a good move if you think so 👍

PR #679, opened 2020-11-24
fix typos/inconsistencies in chapter 2

azzaea:


jennybc:
I'm afraid the entire `fbind()` / foofactors example has had to be replaced. Thanks for this regardless.

PR #680, opened 2020-11-24
add a note on usethis package for testing

azzaea:
The documentation previously suggests `use_testthat()` function is part of `testthat`. However, it is part of `usethis`


jennybc:
I think the current text is correct.

`library(devtools)` attaches usethis as well and in this book we emphasize `library(devtools)` as the way to make all of devtools and usethis available to yourself in an R session.

And we really do want to say that we're using **testthat** as our unit testing package (not usethis).

PR #691, opened 2021-01-31
Update git.Rmd

xiaochi-liu:
It seems that this step is not needed anymore. RStudio solves everything 😁


jennybc:
This whole chapter is likely to go away, so I'm not going to process small tweaks to it:

https://github.com/hadley/r-pkgs#readme

PR #695, opened 2021-02-19
Add comment about `LazyDataCompression`

mrcaseb:
Using a compression method that is not `gzip`, then this should also be specified in the corresponding DESCRIPTION field `LazyDataCompression`, because otherwise the default value `gzip` is used in the installed lazy-loading database. 

Unlike `sysdata.rda`, `xz` is not automatically used for larger exported data, so one should explicitly specify the method.

See 'Writing R Extensions' §1.1.6

### DIFF
--- a/data.Rmd
+++ b/data.Rmd
@@ -143,3 +143,6 @@ Generally, package data should be smaller than a megabyte - if it's larger you'l
    If you've lost the code for recreating the files, you can use 
    `tools::resaveRdaFiles()` to re-save in place.
 
+3. Add the field `LazyDataCompression` to the `DESCRIPTION` file with the optimal
+   compression method, e.g. `LazyDataCompression: xz`.
+


hadley:
Can you provide some evidence this is needed? I've never used it.

mrcaseb:
It's probably only relevant if someone tries to get a package on CRAN that exceeds the 5MB limit. 

I recently released the data only package [fastrmodels](https://cran.r-project.org/package=fastrmodels). 
In v1.0.1 the data folder had 16MB whereas `data/Rdata.rdb` took 20MB. Prof. Ripley sent me a mail today asking me to fix this by using `LazyDataCompression` (I'll quote the mail partly below. There were other packages mentioned in that mail I have removed)

> We see
   20M fastrmodels/data/Rdata.rdb
which are much larger than the data directories:
16M    fastrmodels/data
Please do re-read 'Writing R Extensions' §1.1.6 and check whether you 
are should be using LazyDataCompression with a non-default setting (most 
likely the same as you used for the data directory which in all cases is 
not gzip) or if that does not help, do not use LazyData.
Please correct before 2021-03-19 to safely retain your package on CRAN.
-- 
Brian D. Ripley,                  ripley@stats.ox.ac.uk
Emeritus Professor of Applied Statistics, University of Oxford

I had the data compressed with `xz` but didn't use `LazyDataCompression` because I didn't know it existed tbh.

It's obviously a special usecase with big data but it's probably only this usecase where people read the relevant chapter in your book at all.

hadley:
https://cran.rstudio.com/doc/manuals/r-devel/R-exts.html#Data-in-packages says:

> Some packages using ‘LazyData’ will benefit from using a form of compression other than gzip in the installed lazy-loading database. This can be selected by the --data-compress option to R CMD INSTALL or by using the ‘LazyDataCompression’ field in the DESCRIPTION file. Useful values are bzip2, xz and the default, gzip. The only way to discover which is best is to try them all and look at the size of the pkgname/data/Rdata.rdb file.

But I don't understand this because (e.g.) nycflights13 has:

```R
 tools::checkRdaFiles("data/")
                      size ASCII compress version
data//airlines.rda     386 FALSE     gzip       2
data//airports.rda   46365 FALSE     gzip       2
data//flights.rda  4299947 FALSE    bzip2       2
data//planes.rda     21142 FALSE     gzip       2
data//weather.rda   149212 FALSE    bzip2       2
```

And does not use the `LazyDataCompression` field

hadley:
Ooooh, but this complaint is about the lazy loaded database, which is generated when the package is _built_.

mrcaseb:
> Ooooh, but this complaint is about the lazy loaded database, which is generated when the package is _built_.

That's correct. Sorry I should have clarified this initially. It took me a while to even find the file `data/Rdata.rdb`.

mrcaseb:
For comparison, here are the [check results](https://cran.r-project.org/web/checks/check_results_fastrmodels.html) of the above mentioned package. 
v1.0.2 uses `LazyDataCompression: xz`
v1.0.1 does not. 
image

mrcaseb:
I think the current dev version of R creates a Warning if `LazyDataCompression` is missing for >5MB lazy DBs so this might get more relevant in the future because package maintainers will be asked to fix that.

jennybc:
My current guess is that, if we get into this matter at all, it should appear in the data chapter (not metadata / DESCRIPTION).

jennybc:
My head is in this space now and I still feel like this is a detail that affects relatively few package developers. I'm not denying your experience @mrcaseb. But I'm trying to impose a uniform "level of detail" filter across a wide array of topics, because we only have finite "space". For now, I think the right call is to not delve into this level of detail.

mrcaseb:
I understand @jennybc! Honestly I completely forgot this PR existed. And I agree this may affect only a handful of developers

PR #696, opened 2021-02-25
Capitalize GitLab and Bitbucket

erictleung:


jennybc:
This chapter is likely to go away in the next edition, which is while this is going to sit here and maybe not get merged. But it's a good reminder to globally search for these terms eventually and check capitalization.

PR #699, opened 2021-03-10
Suggest alternative error handling functions

azzaea:
There is no `error` function in `Rcpp`. Probably either `Rcpp::stop()` or `Rf_error()` are accepted by CRAN though?


jennybc:
The content on compiled code is slated to become more minimal and link to external resources, since this is a topic that really demands its own, thorough treatment. So I'm closing PRs that act on the existing text, but thanks.

PR #704, opened 2021-03-30
Suggest more adequate example of multiple line tag

alperceire:
`examples` seems to be a better example than the `example` :)


jennybc:
Thanks!

PR #714, opened 2021-06-08
Test PR CI

Robinlovelace:


jennybc:
Thanks, but I'm OK with the current text AND this content is slated for a thorough revision in any case. I.e. small tweaks to the existing text are kind of moot at this point.

PR #716, opened 2021-06-24
Fix minor grammar problem

PabRod:
If I'm not wrong, all the bulletpoints are subordinates of _"This will"_, and should be conjugated in the third person of the singular (i.e.: this will create, this will add, ...).


hadley:
Thanks!

PR #721, opened 2021-07-20
Fixed a typo.

mjkanji:
Made the following change:

> This tests that fbind() gives an expected result when combining two ~factors~ character vectors and a character vector and a factor.

`x` and `y` are character vectors, not factors.


jennybc:
Thanks, but I'm afraid that chapter had to be rewritten with a new example, due to changes in R itself.

PR #727, opened 2021-09-03
Broken link for Philip Guo link

pdwaggoner:
Broken link to Philip Guo's Basic Unix-like command line tutorial videos. Couldn't find a good alternative for it, so I suggested just omitting it.


jennybc:
Thanks, but the Git content here is going to be scaled back and link to external resources. So it doesn't make sense to groom it at this point. This will all go away before too much longer.

pdwaggoner:
Cool, thanks for looking

PR #738, opened 2021-10-04
fix outdatd information: devtools -> usethis

daattali:


jennybc:
Thanks for this, but the whole chapter on DESCRIPTION has been re-written now.

https://r-pkgs.org/description.html

PR #741, opened 2021-11-02
Also use the css for bs4_book

jennybc:
@Hadley I did this because the switch to `bs4_book()` meant that our special "tips" (for RStudio, a general tip) were no longer being formatted specially.

But now that I look harder at `style.css`, I'm not sure this is really quite the right or complete solution. Should we split the tip stuff out? I also have no idea why some of the tip stuff is under a comment that says `Sidebar formating`. And I see some duplication in the file too. Basically it feels a bit crufty. Thoughts?

### DIFF
--- a/_output.yml
+++ b/_output.yml
@@ -1,4 +1,5 @@
 bookdown::bs4_book:
+  css: style.css
   theme:
     primary: "#982a31"
     fg: "#2b2121"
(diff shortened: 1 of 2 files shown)


hadley:
I think you could probably delete everything in `styles.css` except for the stuff related to the tips section (which is incorrectly labelled sidebars).

jennybc:
I've got the tips back:

Screen Shot 2021-11-02 at 2 55 33 PM
Screen Shot 2021-11-02 at 2 55 50 PM

PR #742, opened 2021-11-06
Replace foofactors and fbind with regexcite and str_split_one()

### LINKED ISSUE #709: Replacing `fbind` with another example
According to [this blog post](https://www.r-bloggers.com/2021/05/new-features-in-r-4-1-0/), one of the new features of R 4.1.0 is that combining two factors now results in another factor instead of an integer vector.

Chapter 2 of this e-book involves creating and packaging a function called `fbind()` which deals with what happens when you tried to concatenate two factors prior this latest major release (outlined in [2.6](https://r-pkgs.org/whole-game.html#write-the-first-function)). Since this behaviour won't be replicated for those following along the book on R >=4.1.0, I think this particular "case study" should be updated.

Sorry if this is already being taken care of by the authors.

jennybc:
Closes #709

TODOs:

- [x] Get syntax highlighting (back) for usethis's messaging. Was it premature to remove the `:::downlit` divs? Or is there something else I need to turn on?
- [x] Get `test()` chunk to actually report something
- [x] The method I'm using of storing code I need to write to file AND include here is suboptimal. It's not syntax highlighted. And the "copy to clipboard" feature doesn't work for the chunks where you are most likely to want it.

I wouldn't say the resolution of these issues is truly elegant (yet), but we're basically getting the desired rendered result. Maybe some of the mechanics can be revisited later.

Some notes re: why this is hard.

In several cases, we can't reveal our actual code, so `eval = FALSE` for the chunk the user sees and should execute and `echo = FALSE` for the chunk we actually execute. `create_package()` is a great example, due to the fact that we have to write to a temporary file path. But `echo = FALSE` seems to mean the chunk is no longer treated as source code (and output), so then you also lose highlighting. Which is why we have to use explicit `:::downlit` divs.

In other cases, we have code or other text that we want to write to file, in the toy package, and display to the user. And we want that also to be perceived as source code, so that we get the copy button. We may or may not want syntax highlighting, depending on whether it's DESCRIPTION (no) or LICENSE (no) or `.R` (yes). This requires specific methods of handling the associated text.

Finally, it's proving to be quite hard to capture the result of `test()`, as the user will see in their interactive session, with color and emoji and suppressing progress reporting.


Whole-game.Rmd
```diff
@@ -54,51 +60,61 @@ Do you have an old version of devtools? Compare your version against ours and up
...
+  - This is completely optional in your work, but highly recommended.
+    You'll see how Git and GitHub help us expose all the intermediate stages of
+    our toy package.
+* Access to established workflows for installation, getting help, and checking quality.
+  - Documentation for individual functions via [roxygen2](https://CRAN.R-project.org/package=roxygen2).
+  - Unit testing with [testthat](https://testthat.r-lib.org).
+  - Documentation for the package as a whole via an executable `README.Rmd`.
 
-We call the package __foofactors__ and it will have a couple functions for handling factors. Please note that these functions are super simple and definitely not the point! For a proper package for factor handling, please see [forcats](https://forcats.tidyverse.org).
+We call the package **regexcite** and it will have a couple functions that make common tasks with regular expressions easier.
```
jennybc, line 77: It doesn't really have a "couple" functions, it's just one. But I still think this wording is OK. Or, if we started to suggest exercises for the reader, we could certainly suggest another function or two for them to practice adding to the package (complete with roxygen comment, tests, inclusion in NAMESPACE, mention in README, etc.).

Whole-game.Rmd
```diff
@@ -109,300 +125,375 @@ create <- !is.na(where)
...
 ```
 
-:::downlit
-```{r create-package, eval = create, echo = debug, R.options = list(usethis.description = NULL)}
-create_package(foopath, open = FALSE, rstudio = TRUE)
-proj_set(foopath)
+```{r create-package, eval = create, echo = debug}
+withr::with_options(
+  list(usethis.description = NULL),
+  create_package(pkgpath, open = FALSE, rstudio = TRUE)
```
jennybc, line 150: usethis's messaging is not rendered in color, so I've clearly turned too many things off or they need to be revised for the current downlit world.

Whole-game.Rmd
```diff
@@ -109,300 +125,375 @@ create <- !is.na(where)
...
+Often you know that your input is morally a scalar, i.e. it's just a single string, and really want the output to be the character vector its parts.
 
-Huh? Many people do not expect the result of catenating two factors to be an integer vector consisting of the numbers 1, 2, 3, and 4. What if we coerce each factor to character, catenate, then re-convert to factor?
+This leads R users to employ various methods of "unlist"-ing the result:
 
 ```{r}
-factor(c(as.character(a), as.character(b)))
+unlist(strsplit(x, split = ","))
+
+strsplit(x, split = ",")[[1]]
```
jennybc, line 286: I am half-seriously considering doing some sort of search to grab lines containing `strsplit()` in R code on GitHub and do a rough analysis of what proportion are one of these forms. I feel like I've even done this in the past and the proportion is quite high! Would be good for a footnote at least.

Whole-game.Rmd
```diff
@@ -537,208 +648,327 @@ cat(pretty_install(), sep = "\n")
...
 
 Going forward, your tests will mostly run *en masse* and at arm's length via `test()`:
 
-*TODO: work on the aesthetics of this output. Or maybe testthat 3e will save me the trouble.*
+<!-- TODO: work on the aesthetics of this output. Or maybe testthat 3e will save me the trouble. -->
 
-```{r test, eval = create, message = FALSE, R.options = list(testthat.default_reporter = testthat::ProgressReporter$new(update_interval = Inf))}
+```{r test}
+#| eval = create, message = FALSE,
+#| R.options = list(testthat.default_reporter = testthat::ProgressReporter$new(update_interval = Inf))
```
jennybc, line 730: This is not achieving what we want. Currently this chunk shows absolutely nothing.

jennybc: What it should show is more like:

``` r
test()
#> ℹ Loading regexcite
#> ℹ Testing regexcite
#> ✓ | F W S  OK | Context
#> ⠏ |         0 | str_split_one                                                      ✓ |         4 | str_split_one
#> 
#> ══ Results ════════════════════════════════════════════════════════════════════════
#> [ FAIL 0 | WARN 0 | SKIP 0 | PASS 4 ]
#> 
#> 😀
```

Whole-game.Rmd
jennybc: Compare how this chunk looks in the preview of this PR to how it currently looks in the main site:
https://r-pkgs.org/whole-game.html#create_package

Whole-game.Rmd
jennybc: Informally, if I do a GitHub search for "strsplit path:R user:cran", 9 of the 10 hits on the first page use either `unlist()` or `[[1]]`. The remaining one uses `[[1]]` when the result is actually used in the next line. So 10 for 10 of these examples actually want the un-listed result 🤣

Whole-game.Rmd
jennybc: I've put all the `:::downlit` divs back. It seems they are still necessary for a chunk with `echo = FALSE`.

hadley: Oh yeah, that must be why they edited them in the first place.

PR #743, opened 2021-11-11
Add `r ''` to the chunk header instead of inserting a zero-width space

yihui:
Doc: https://bookdown.org/yihui/rmarkdown-cookbook/verbatim-code-chunks.html

This fixes the warning that @jennybc asked about:

```
Warning message:
In xfun::prose_index(x) : Code fences are not balanced
```

Hopefully in the next version of knitr, it will be less hacky to write verbatim code chunks.


jennybc:
Thanks!

PR #747, opened 2021-11-20
Fix typo in Whole-game.Rmd

henningte:
I've removed a duplicate "the" (typo):

> The very best way to render `README.Rmd` is with `build_readme()`, because it takes care to render with the ~~the~~ most current version of your package, i.e. it installs a temporary copy from the current source.


jennybc:
Thanks!

PR #751, opened 2021-12-16
Prefer `.rda` over `.RData` ext for exported data

salim-b:
`usethis::use_data(..., internal = FALSE)` creates `.rda` files, so this is more consistent.


jennybc:
Thanks!

Issue #526, opened 2019-02-05
Remove most of the Git/GitHub content and link to Happy Git


jennybc:
Figure out how to leave some motivating / mission critical Git/GitHub content here (where?), but then link out to <https://happygitwithr.com>.

PR #752, opened 2021-12-22
Update git.Rmd

PursuitOfDataScience:
Added a quick note containing a few shell command lines for Windows users.


jennybc:
Thanks, but this entire chapter is largely going away in the 2nd edition, so I'm not merging PRs against it at this point. The revision process is still underway.

https://github.com/hadley/r-pkgs/issues/526

PR #753, opened 2021-12-22
Update git.Rmd

PursuitOfDataScience:
Philip Guo's videos are gone as the links are expired, and there is no way for me to retrieve the new links on the Internet. I would recommend to delete the line. Other links work fine as of today.


jennybc:
Duplicate of #727

PR #754, opened 2021-12-30
Update Introduction.Rmd

PursuitOfDataScience:
Since the book is formally written by more than one author, I think it is appropriate to use "we" rather than "I" throughout the book. I only "corrected" a few places in this section only. Hopefully this is something authors can be mindful about when moving to the third edition.


jennybc:
The introduction will probably be one of the last things we revise. I can live with it like this for the time being.

PR #759, opened 2022-02-01
Update Structure.Rmd

PursuitOfDataScience:
Fixed a typo.


jennybc:
This is intentional. There's a tradition of talking about R users as useRs.

PR #760, opened 2022-02-01
Update check.Rmd

PursuitOfDataScience:
It is actually `Author`, not `Authors`. I had `Author` but changed to `Authors` after following the book and then `CMD Check` threw me an error (attached below).

image


jennybc:
This is touched on in the revised chapter on DESCRIPTION and this content is going to be heavily reworked anyway.

https://github.com/hadley/r-pkgs/blob/43d2882ead99879cf6a9b3d1c29f9107b2ed687a/Metadata.Rmd#L605-L606

PR #762, opened 2022-02-06
`expect_is()` is superseded, new options added

alesantuz:
As mentioned [here](https://rdrr.io/cran/testthat/man/expect_is.html), the new [inheritance-expectations](https://rdrr.io/cran/testthat/man/inheritance-expectations.html) should be used instead of `expect_is()`.

### DIFF
--- a/testing.Rmd
+++ b/testing.Rmd
@@ -225,7 +225,9 @@ While you'll normally put expectations inside tests inside files, you can also r
     expect_error(1 / 2) 
     ```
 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks! You've pointed out something in a chapter we haven't revised yet for the 2nd edition and it looks like this will be easier to fix as part of the holistic rethink of this content. I'm closing this, but appreciate the pull request.

PR #764, opened 2022-02-07
Update check.Rmd

PursuitOfDataScience:
Fixed a "typo".


jennybc:
This is in a chapter that is clearly going to be rewritten, as it the content re: Git. And I'm sure we have other inconsistencies around git vs Git that should all be settled at once. Thanks but just closing this one.

PR #769, opened 2022-02-16
explicit call to package because no packages loaded in new session

michaelboerman:
The previous step, `create_package`, opens a new session. No packages are loaded in this new session, so calling `use_git()` fails. I propose either referencing the package via `usethis::use_git()`, or, more robustly (for later calls), to include `library(devtools)` and a brief explanation.


jennybc:
Thanks! I'd rather mention this as the conclusion of the previous section (re: `create_package()`) and call attention to it with a sentence. So I modified your PR to implement your suggestion, my way 😁

PR #771, opened 2022-02-19
Update check.Rmd

PursuitOfDataScience:
Added an extra feedback from CRAN to me.


jennybc:
This advice is already presented in the revised chapter on DESCRIPTION.

https://github.com/hadley/r-pkgs/blob/43d2882ead99879cf6a9b3d1c29f9107b2ed687a/Metadata.Rmd#L476-L478

PR #774, opened 2022-03-13
Correct data.Rmd

pwaeckerle:
Changing the proposed path to store raw data in the introduction, to be compliant with `usethis::use_data_raw()`


jennybc:
The `data-raw` folder is not where the data should be stored, but rather where we keep the code that converts raw wild-caught data into whatever is shipped in the package.

PR #777, opened 2022-04-20
Update testing chapter

### LINKED ISSUE #477: More in-depth treatment of testing
As updates are considered for a second version: It would be really helpful to expand the excellent introduction to testing to cover best practices for more complex tests, such as how to store objects that should be used in >1 test script.

### LINKED ISSUE #483: Usual Suspects for specific failure modes
Tests that pass during `devtools::test()` but then fail during `devtools::check()`

One cause: files present in source package but that match one of the criteria for auto-Rbuildignore and, therefore, aren't present when tests run at the end of `R CMD check`.

> In addition, directories from source control systems or from eclipse, directories with names ending `.Rcheck` or `Old` or `old` and files `GNUMakefile`, `Read-and-delete-me` or with base names starting with `‘.#’`, or starting and ending with `‘#’`, or ending in `‘~’`, `‘.bak’` or `‘.swp’`, are excluded by default.

from <https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Building-package-tarballs>

### LINKED ISSUE #551: Emphasise that tests should be concrete
Along the lines of https://mtlynch.io/good-developers-bad-tests/

### LINKED ISSUE #556: Talk about setup / teardown and helper files
Also emphasize that in most cases code should go in these rather than in `tests/testthat.R`, since the latter is not executed with `devtools::test()`.

This is a pretty common source of issues, for two recent examples,
  - https://github.com/r-lib/devtools/issues/2038
  - https://github.com/hadley/r-pkgs/issues/483#issuecomment-480200246

But I could dig up more as well if needed.

### LINKED ISSUE #580: How to make dependencies available in tests
General observation, sharpened by working on tidyr revdeps: would be good to clarify how one should call functions in other packages in code below `tests/`. Address this for packages that appear in Imports vs Suggests. Address for code re: helpers & setup vs. inside a test.

### LINKED ISSUE #711: Testing in presence/absence of suggested packages
Discuss this situation and various solutions.

jennybc:
Closes #477, closes #483, closes #551, closes #556, closes #580, closes #711

### DIFF
--- a/Testing.Rmd
+++ b/Testing.Rmd
@@ -30,13 +30,11 @@ While you _are_ testing your code in this workflow, you're only doing it informa
 The problem with this approach is that when you come back to this code in 3 months time to add a new feature, you've probably forgotten some of the informal tests you ran the first time around.
 This makes it very easy to break code that used to work. 
 
-<!-- This next paragraph is very "I" based. But it's also hard to rephrase. Let's discuss. -->
-
-I started using automated tests because I discovered I was spending too much time re-fixing bugs that I'd already fixed before.
-While writing code or fixing bugs, I'd perform interactive tests to make sure the code worked.
-But I never had a system which could store those tests so I could re-run them as needed.
-I think that this is a common practice among R programmers.
-It's not that you don't test your code, it's that you don't automate your tests.
+Many of us embrace automated testing when we realize we're re-fixing a bug for the second or fifth time.
+While writing code or fixing bugs, we might perform some interactive tests to make sure the code we're working on does what we want.
+But it's easy to forget all the different use cases you need to check, if you don't have a system for storing and re-running the tests.
+This is a common practice among R programmers.
+The problem is not that you don't test your code, it's that you don't automate your tests.
 
 In this chapter you'll learn how to transition from informal *ad hoc* testing, done interactively in the console, to automated testing (aka unit testing).
 While turning casual interactive tests into formal tests requires a little more work up front, it pays off in four ways:
@@ -224,7 +222,7 @@ When you are rapidly iterating on a function, you might work at the level of ind
 As the code settles down, you'll run entire test files and eventually the entire test suite.
 
 **Micro-iteration**: This is the interactive phase where you initiate and refine a function and its tests in tandem.
-Here you will run `load_all()` often, and then execute individual expectations or whole tests interactively in the console.
+Here you will run `devtools::load_all()` often, and then execute individual expectations or whole tests interactively in the console.
 Note that `load_all()` attaches testthat, so it puts you in the perfect position to test drive your functions and to execute individual tests and expectations.
 
 ```{r, eval = FALSE}
--- a/Testing.Rmd
+++ b/Testing.Rmd
@@ -1275,6 +1275,19 @@ Sometimes you need even more control over the file name, in which case two techn
   the exact name you provide.
 * Use `withr::local_tempdir()` to create a self-deleting temporary directory and
   write intentionally-named files inside this directory.
+  
+### Files ignored by testthat
+
+testhat only automatically executes files where these are both true:
+
+* File is a direct child of `tests/testthat/`
+* File name starts with one of the specific strings:
+  - `helper`
+  - `setup`
+  - `test`
+
+It is fine to have other files or directories in `tests/testthat/`, but testthat won't automatically do anything with them (other than the `_snaps` directory, which holds snapshots).
+You can keep other files here, for reading or copying, and use `testthat::test_path()` to build filepath.
 
 ## Test fixtures
 
@@ -1366,7 +1379,7 @@ test_that("foofy() does that", {
 
 Where should the `local_useful_thing()` helper be defined?
 All the advice given above for `new_useful_thing()` applies:
-define it below `R/` or a test helper file.
+define it below `R/` or in a test helper file.
 
 To learn more about writing custom helpers like `local_useful_thing()`, see the [testthat vignette on test fixtures](https://testthat.r-lib.org/articles/test-fixtures.html).
 
@@ -1408,7 +1421,7 @@ Now we can revisit a file listing from earlier, which addressed exactly this sce
 This shows static test files stored in `tests/testthat/fixtures/`, but also notice the companion R script, `make-useful-things.R`.
 From data analysis, we all know there is no such things as a script that is run only once.
 Refinement and iteration is inevitable.
-This also holds true to test objects like `useful_thing1.rds`.
+This also holds true for test objects like `useful_thing1.rds`.
 We highly recommend saving the R code used to create your test objects, so that they can be re-created as needed.
 
 ## Building your own testing tools
--- a/Testing.Rmd
+++ b/Testing.Rmd
@@ -1,16 +1,9 @@
 <!--
 Thought dump re: revision, stuff I haven't added yet (and maybe never will?)
 
-Mentality: you're visiting a wild, uninhabited nature preserve.
-Bring your own supplies.
-Leave everything as you found it.
-Practice conservatism, minimalism.
-
-
 
 mocking, giving yourself a way to exercise code in various "as if" scenarios
   
-
 
 When does our file naming / organisation  advise not work? sometimes a package has one main function and following the convention would result in having one monster file of tests.
 
@@ -30,18 +23,10 @@ namespace and scope thoughts:
   same below R/
 
 
-skipping
-
-
 web, APIs, auth, secrets
 
 
 CRAN specific matters
-
-Why Good Developers Write Bad Unit Tests
-https://mtlynch.io/good-developers-bad-tests/
-"Your natural inclination might be to delegate all the uninteresting code to test helper methods, but you should first ask a more vital question: why is the system so difficult to test? Excessive boilerplate code is often a symptom of weak architecture."
-"When tempted to write test helper methods, try refactoring your production code instead."
 -->
 
 # Testing {#tests}
@@ -164,7 +149,11 @@ This will:
 1.  Create a `tests/testthat/` directory.
 
 1.  Add testthat to the `Suggests` field in the `DESCRIPTION`.
-    Specifies testthat 3e in the `Config/testthat/edition` field.
+    Specify testthat 3e in the `Config/testthat/edition` field.
+    Here are what the relevant `DESCRIPTION` fields might look like:
+    
+    Suggests: testthat (>= 3.0.0)
+    Config/testthat/edition: 3
 
 1.  Create a file `tests/testthat.R` that runs all your tests when
     `R CMD check` runs. (You'll learn more about that in 
@@ -175,7 +164,7 @@ This initial setup is something you do once per package.
 The `tests/testthat.R` script is something you should generally regard as fixed.
 It is run during `R CMD check` (and, therefore, `devtools::check()`), but is not used in other test-running scenarios (such as `devtools::test()` or `devtools::test_active_file()`).
 If you want to do something that affects all of your tests, there is usually a better mechanism than modifying the standard `tests/testthat.R` script.
-You'll see this below when we discuss mechanisms for test setup / teardown and test helper files.
+You'll see this below when we discuss different types of test helpers and mechanisms for setup / teardown.
 
 ### Create a test
 
@@ -714,6 +703,7 @@ test_that("thingy exists", {
 exists("thingy")
 ```
 
+The `thingy` object lives and dies entirely within the confines of `test_that()`.
 However, testthat doesn't know how to cleanup after actions that affect other aspects of the R landscape:
 
 * The filesystem: creating and deleting files, changing the working directory,
@@ -723,7 +713,7 @@ However, testthat doesn't know how to cleanup after actions that affect other as
 
 If it's easy to avoid making such changes in your test code, that is the best strategy.
 But if it's unavoidable, then you have to make sure that you clean up after yourself.
-This mindset is very similar to one we advocated for in section \@ref(#code-r-landscape), when discussing how to design well-mannered functions.
+This mindset is very similar to one we advocated for in section \@ref(code-r-landscape), when discussing how to design well-mannered functions.
 
 When possible, make each test self-sufficient:
 
@@ -737,9 +727,10 @@ In the testing domain, this is sometimes referred to as *setup and teardown* and
 
 But first, let's connect this back to very similar advice offered in section \@ref(code-r-landscape), which covers your package's "main code", i.e. everything below `R/`:
 
-The `.R` files below `R/` should consist almost entirely of function definitions.
-Any other top-level code is suspicious and should be carefully reviewed for possible conversion into a function.
-Similarly, the `test-*.R` files below `tests/testthat/` should consist almost entirely of calls to `test_that()`.
+> The `.R` files below `R/` should consist almost entirely of function definitions.
+> Any other top-level code is suspicious and should be carefully reviewed for possible conversion into a function.
+
+Similarly, `tests/testthat/test-*.R` files should consist almost entirely of calls to `test_that()`.
 Any other top-level code is suspicious and should be carefully reviewed for conversion to a more official method of achieving whatever its goal is.
 
 Two concrete examples:
@@ -747,7 +738,8 @@ Two concrete examples:
 * `source()` should not appear in your tests.
   There are several special files with an official role in testthat workflows
   (see below), not to mention the entire R package machinery, that provide
-  betterways to make functions, objects, and other logic available in your tests.
+  better ways to make functions, objects, and other logic available in your
+  tests.
 * `library()` should (almost certainly) not appear in your tests.
   Similar to advice for the code below `R/`, it's usually better to access
   functions from a dependency via `otherpkg::fun()`.
@@ -929,7 +921,8 @@ test_that("foofy() does that", {
 Where should the `new_useful_thing()` helper be defined?
 For new code and actively maintained code, we recommend defining it like any other internal helper, i.e. in a file below `R/`.
 If it logically fits into an existing `.R` file, put it there.
-Otherwise, consider making a file for test utilities, such as `R/utils-testing.R`.
+Otherwise, consider making a file for test utilities, such as `R/utils-testing.R` or `R/test-helpers.R`.
+
 A key advantage of this solution is that the `new_useful_thing()` helper is available to you during interactive development after `devtools::load_all()` and during automated test runs (`devtools::test_active_file()`, `devtools::test()`, etc.), because this is true of all unexported functions.
 Helper code that is lying around in other places, e.g. sprinkled around in test files, does not have this property.
 
@@ -1025,8 +1018,9 @@ Setup files are run by testthat before running whole test files.
 Note that setup code is *not* run by `devtools::load_all()`.
 
 If any of your setup should be reversed after test execution, you should also include the necessary teardown code in `setup.R`.
-We recommend maintaining teardown code with the setup code, in `setup.R`, which makes it easier to ensure they stay in sync.
-A legacy approach which still works, but is no longer recommended, is to put teardown code in `tests/testthat/teardown.R`.
+We recommend maintaining teardown code alongside the setup code, in `setup.R`, because this makes it easier to ensure they stay in sync.
+The artificial environment `teatdown_env()` exists as a magical handle to use in `withr::defer()` and `withr::local_*()` / `withr::with_*()`.
+A legacy approach (which still works, but is no longer recommended) is to put teardown code in `tests/testthat/teardown.R`.
 
 Here's a `setup.R` example from the reprex package, where we turn off clipboard and HTML preview functionality during testing:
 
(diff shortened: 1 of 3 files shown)


Testing.Rmd
```diff
@@ -97,87 +9,254 @@ It ensures that your code does what you want it to do.
...
+    Suggests: testthat (>= 3.0.0)
+    Config/testthat/edition: 3
 
 1.  Create a file `tests/testthat.R` that runs all your tests when
     `R CMD check` runs. (You'll learn more about that in 
     [automated checking](#r-cmd-check).)
+    
+This initial setup is something you do once per package.
+
+The `tests/testthat.R` script is something you should generally regard as fixed.
```
hadley, line 133: I think you could be ever stronger. "DON'T TOUCH THIS"

Testing.Rmd
```diff
@@ -97,87 +9,254 @@ It ensures that your code does what you want it to do.
...
+It is run during `R CMD check` (and, therefore, `devtools::check()`), but is not used in other test-running scenarios (such as `devtools::test()` or `devtools::test_active_file()`).
+If you want to do something that affects all of your tests, there is usually a better mechanism than modifying the standard `tests/testthat.R` script.
+You'll see this below when we discuss different types of test helpers and mechanisms for setup / teardown.
+
+### Create a test
+
+As you define functions in your package, in the files below `R/`, you add the corresponding tests to `.R` files in `tests/testthat/`.
+We recommend that the organisation of test files match the organisation of `R/` files, discussed in section \@ref(code-organising):
```
hadley, line 141:
```suggestion
We strongly recommend that the organisation of test files match the organisation of `R/` files, discussed in section \@ref(code-organising):
```

Testing.Rmd
```diff
@@ -97,87 +9,254 @@ It ensures that your code does what you want it to do.
...
+```
+
+The `use_r()` / `use_test()` duo has some convenience features that make them "just work" in many common situations:
+
+* When determining the target file, they can deal with the presence or absence
+  of the `.R` extension and the `test-` prefix.
+  - Equivalent: `use_r("foofy.R")`, `use_r("foofy")`
+  - Equivalent: `use_test("test-blarg.R")`, `use_test("blarg.R")`, `use_test("blarg")`
+* If the target file already exists, it is opened for editing. Otherwise, the
+  target is created and then opened for editing.
```
hadley, line 177: It feels like this might be just a bit much detail.

Testing.Rmd
```diff
@@ -97,87 +9,254 @@ It ensures that your code does what you want it to do.
...
+**Mezzo-iteration**: As one file's-worth of functions and their associated tests start to shape up, you will want to execute the entire file of associated tests, perhaps with `testthat::test_file()`:
+
+```{r, eval = FALSE}
+testthat::test_file("tests/testthat/test-foofy.R")
+```
+
+:::rstudio-tip
+In RStudio, you don't even have to specify the full filepath!
+`devtools::test_active_file()` infers the target test file from the file you are actively editing, similar to how `use_r()` and `use_test()` work.
+The easiest way to invoke this is via "Run a test file" in the Addins menu or with the keyboard shortcut Ctrl/Cmd + T.
```
hadley, line 234: You do need to set up that keyboard shortcut yourself

Testing.Rmd
```diff
@@ -207,65 +286,93 @@ A test failure looks something like this:
...
+The idea is that you'll modify your code (either the functions defined below `R/` or the tests in `tests/testthat/`) until all tests are passing.
 
 ## Test structure {#test-structure}
 
 A test file lives in `tests/testthat/`.
 Its name must start with `test`.
-Here's an example of a test file from the stringr package:
+We will inspect and execute a test file from the stringr package.
+
+But first, for the purposes of rendering this book, we must attach stringr and testthat.
```
hadley, line 298: Should we add a block that is "we're running this code so the book work, but you should never do this"?

Testing.Rmd
```diff
@@ -207,65 +286,93 @@ A test failure looks something like this:
...
+  specific kind of error.
+
+Tests are organised hierarchically:
+__expectations__ are grouped into __tests__ which are organised in __files__:
+
+* A __file__ holds multiple related tests.
+  In this example, the file `tests/testthat/test-dup.r` has all of the tests
+  for `stringr::str_dup()`, which is defined in `R/dup.r`.
+  A test failure is always reported by file name, which is why the file name
+  should convey the general context, e.g. the function that is being tested.
```
hadley, line 352:
```suggestion
  In this example, the file `tests/testthat/test-dup.r` has all of the tests
  for the code in `R/dup.r`.
```

Testing.Rmd
```diff
@@ -275,140 +382,640 @@ All expectations have a similar structure:
...
     expect_equal(10, 10)
+    expect_equal(10, 10L)
     expect_equal(10, 10 + 1e-7)
     expect_equal(10, 11)
     ```
   
-    If you want to test for exact equivalence, or need to compare a more
-    exotic object like an environment, use `expect_identical()`.
+  If you want to test for exact equivalence or need to compare more exotic
+  objects like environments, use `expect_identical()`.
```
hadley, line 411:
```suggestion
  If you want to test for exact equivalence, use `expect_identical()`.
```

Testing.Rmd
```diff
@@ -275,140 +382,640 @@ All expectations have a similar structure:
...
     ```
 
-*   `expect_match()` matches a character vector against a regular expression. 
-    The optional `all` argument controls whether all elements or just one 
-    element needs to match.
-    This is powered by `grepl()` (additional arguments like
-    `ignore.case = FALSE` or `fixed = TRUE` are passed on down).
-
-    ```{r, error = TRUE, eval = FALSE}
+* `expect_match()` matches a character vector against a regular expression.
```
hadley, line 420: I think this should come later or maybe not at all?

Maybe there should be (or already is) a section that lists a bunch of shortcuts, e,g:

* `expect_length(x, n)` is a short cut for `expect_equal(length(x), n)`
* `expect_match(x, pattern)` is a shortcut for `expect_true(all(grepl(pattern, n)))` 

etc

Testing.Rmd
```diff
@@ -275,140 +382,640 @@ All expectations have a similar structure:
...
-*   Four variations of `expect_match()` let you check for other types of 
-    result: `expect_output()`, inspects printed output; `expect_message()`,
-    messages; `expect_warning()`, warnings; and `expect_error()` errors.
-    
-    ```{r, error = TRUE, eval = FALSE}
-    a <- list(1:10, letters)
-
-    expect_output(str(a), "List of 2")
-    expect_output(str(a), "int [1:10]", fixed = TRUE)
+* `expect_s3_class()` and `expect_s4_class()` check that an object `inherit()`s
```
hadley, line 438: I think `expect_equal()` + `expect_error()` (and friends) and `expect_snapshot()` are by far the most important and should come first.

Testing.Rmd
```diff
@@ -275,140 +382,640 @@ All expectations have a similar structure:
...
+  [25] "x" | "x" [24]
+  [26] "y" | "y" [25]
+  [27] "z" | "z" [26]
+           - "X" [27]
+```
+
+There is always a warning upon initial snapshot creation.
+The snapshot is added to `tests/testthat/_snaps/diff.md`, under the heading "side-by-side diffs work", which comes from the test's description.
+The snapshot looks exactly like what a user sees interactively in the console, which is the experience we want to check for.
+The snapshot file is *also* very readable, which is pleasant for the package developer.
```
hadley, line 573: Some comment/footnote about PRs here

Testing.Rmd
```diff
@@ -275,140 +382,640 @@ All expectations have a similar structure:
...
+})
+
+test_that("subtraction works", {
+  expect_equal(5 - useful_thing, 2)
+})
+```
+
+This does sort of work, because when `useful_thing` is not found in the test-specific environment, the search continues in the parent environment, where `useful_thing` will often be found.
+When testthat executes an entire test file, `useful_thing` will get defined and made available for subsequent tests in that file.
+However, during interactive development, there is no gesture to systematically identify and execute such code.
```
hadley, line 825: This is particularly frustrating when the test at line 573 is failing and all the setup code is at line 1.  Or worse when it depends on code smooshed between two other tests at some random location that made sense at the time you wrotei t.

Testing.Rmd
```diff
@@ -275,140 +382,640 @@ All expectations have a similar structure:
...
+Top-level code, outside of any test, is in a sort of No Man's Land and the tests relying on it work more by coincidence, than by definition.
```
hadley, line 826:
```suggestion
Top-level code, outside of any test, is in a sort of No Man's Land: it's fine when your tests are all passing, but when something goes wrong and a test fails you now have to encounter a bunch of extra pain when it need the least.
```

Testing.Rmd
jennybc: https://github.com/hadley/r-pkgs/issues/778

Testing.Rmd
```diff
@@ -452,27 +1097,109 @@ In short, there are times when, instead of getting a failure, you just want to s
...
+test_that("foo api returns bar when given baz", {...})
+
+test_that("foo api returns an errors when given qux", {...})
+```
+
+We have mixed feelings about this practice.
+Within the realm of top-level code in test files, this is probably on the "more acceptable" / "least offensive" end.
+But once a test file does not fit entirely on your screen, it creates an implicit yet easy-to-miss connection between the skip(s) and individual tests.
+
+At the complete other extreme, another practice that can make sense in very specific situations is to build a `skip()` directly into a package function.
```
hadley, line 1128: Would be better in https://testthat.r-lib.org/articles/skipping.html, I think.

Testing.Rmd
```diff
@@ -452,27 +1097,109 @@ In short, there are times when, instead of getting a failure, you just want to s
...
+So pkgdown might as well call `testthat::is_testing(), instead of inlining it.
+
+pkgdown lists testthat in Suggests, not Imports, so that is also interesting to me. Hmmm.
+-->
+
+Advantages of the embedded `skip()` approach: Feels elegant and concise.
+
+Disadvantages of this approach: Seems to violate "Good production code is well-factored; good test code is obvious", i.e. when you're focused on the tests, the existence of this skip is easy to miss.
+
+<!-- I want to talk about the danger of skipping, i.e. when you think everything's "green", but actually you're not running any / many tests, because some pre-condition is not met. -->
```
hadley, line 1178: One challenge with skips is that they are currently completely invisible — if you automatically skip too many tests it's easy to fool yourself that all your tests are passing when in fact they're just being skipped. In an ideal world, your CI/CD would make it easy to see how many tests are being skipped and how that changes over time. 

Good practice to actually dig into the R CMD check results and look at the test results and make sure the skips are as you expect. But this tends to be something you have to learn through experience.

Testing.Rmd
```diff
@@ -452,27 +1097,109 @@ In short, there are times when, instead of getting a failure, you just want to s
...
+
 ### Building your own testing tools
 
-As you start to write more tests, you might notice duplication in your code.
+<!--
```
hadley, line 1182: Counter-proposal: all this code goes into a testthat vignette, and instead focus on a hyperlocal function. Here's an example from stringr:

```R
test_that("truncations work for all sides", {

  trunc <- function(direction) str_trunc(
    "This string is moderately long",
    direction,
    width = 20
  )

  expect_equal(trunc("right"),   "This string is mo...")
  expect_equal(trunc("left"),    "...s moderately long")
  expect_equal(trunc("center"),  "This stri...ely long")
})
```

hadley:
```R
test_that("can fill in missing cells", {
  df <- tibble(g = c(1, 2), var = c("x", "y"), val = c(1, 2))

  widen <- function(...) {
    df %>% pivot_wider(names_from = var, values_from = val, ...)
  }

  expect_equal(widen()$x, c(1, NA))
  expect_equal(widen(values_fill = 0)$x, c(1, 0))
  expect_equal(widen(values_fill = list(val = 0))$x, c(1, 0))
})
```

Testing.Rmd
```diff
@@ -512,80 +1242,118 @@ test_that("floor_date works for different units", {
...
-CRAN will run your tests on all CRAN platforms: Windows, Mac, Linux and Solaris.
+CRAN will run your tests on all CRAN platforms: Windows, Mac, and Linux.
 There are a few things to bear in mind:
 
 * Tests need to run relatively quickly - aim for under a minute.
   Place `skip_on_cran()` at the beginning of long-running tests that shouldn't
   be run on CRAN - they'll still be run locally, but not on CRAN.
 
-* Note that tests are always run in the English language (`LANGUAGE=EN`) and
+* ~Note that tests are always run in the English language (`LANGUAGE=EN`) and
```
hadley, line 1354: This should would be a good place to explain `local_reproducible_output()`

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
   32-bit versions of R) so use `expect_equal()` rather than
   `expect_identical()`.
 
-[tdd]:https://en.wikipedia.org/wiki/Test-driven_development
-[extreme-programming]:https://en.wikipedia.org/wiki/Extreme_programming
+CRAN revision thought dump:
+
+* Enrich description of platforms, i.e. say more than "Windows, Mac, Linux".
+  The check flavors: <https://cran.r-project.org/web/checks/check_flavors.html>.
+  - Draw a line to how this determines which "flavors" a package should be
```
hadley, line 1371: Mention r-hub

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+    checked on.
+    Depends on how broad the user base is and the "trickiness" of the package's
+    code (e.g. does it need compilation?).
+    Do a cost benefit analysis.
+    Connect that to which `usethis::use_github_action_check_*()` variant is
+    appropriate.
+* How to decide whether to let a test run on CRAN.
```
hadley, line 1378: Any test that depends on something that is not under your direct control is a candidate for skipping on CRAN.

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+  If the test can fail WITHOUT indicating a defect in your package that merits
+  removal from CRAN, do not let CRAN run the test. Examples:
+  - Accessing a website or API that could ever be down or have an
+    occasional error (so: any website or API on planet earth).
+  - Exact formatting of a message.
+  - File system work typical of an authentic user that touches files any where
```
hadley, line 1384: If you have to write to a file use `withr::local_temp_file()`. This will protect you against 99% of the most likely problems and if new problems arise we can fix centrally in one location.

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+* How to prevent a test from running on CRAN.
+  - `skip_on_cran()`, `skip_on_os()`
+  - Default behaviour for snapshot tests
+  - Keep in a separate repo
+
+## Miscellaneous sections to (maybe) add
+
+### Mocking
+
+At the very least, I think we need to point to packages that handle this and outline situations when mocking is one of the reasonable approaches.
```
hadley, line 1396: Brief description of what mocking is and a survey of the state of the art currently.

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+
+It's not actually a topic I want to get deeply into, because it feels pretty specialized and I don't think we have detailed, strongly held, stable opinions about it.
+
+### Filesystem matters
+
+Do not write into `tests/testthat/` during your tests.
+
+The only place you should write into is session temp dir.
+
+Even then, best practice is to clean up any files or folder you create.
```
hadley, line 1406: ONLY EVER USE `withr::local_temp_file()`. ALL OTHER PATHS LEAD TO PAIN AND SUFFERING

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+
+### Introspection
+
+`NOT_CRAN`, `skip_on_cran()`
+
+`TESTTHAT`, `is_testing()`
+
+`testing_package()`
+
+### Test coverage
```
hadley, line 1442: I think this should come much earlier since it's now a key part of my workflow.  Would be good in "what to test"

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+
+* What is test coverage?
+* Use covr to learn your test coverage.
```
hadley, line 1445: Talk about binding `Cmd + R` to `devtools::test_coverage_local_file()` and running repeatedly while you're writing your tests.

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+* Use GitHub Actions to keep tabs on your test coverage.
+* Test coverage should be "high", but going from 90% to 100% coverage is not
+  always the best use of your development time and energy.
```
hadley, line 1448: Especially since the last 1% might require excessive gymnastics to cover. Don't sacrifice maintainability for covering some weird edge case that isn't that important.

Testing.Rmd
```diff
@@ -596,5 +1364,101 @@ There are a few things to bear in mind:
...
+  - File system work typical of an authentic user that touches files any where
+    but the R session's temp directory. This includes reading or writing the
+    clipboard.
+* How to prevent a test from running on CRAN.
+  - `skip_on_cran()`, `skip_on_os()`
+  - Default behaviour for snapshot tests
+  - Keep in a separate repo
+
+## Miscellaneous sections to (maybe) add
+
```
hadley, line 1393: Something about random numbers, `withr::local_seed()` etc.

Testing.Rmd
```diff
@@ -207,210 +310,434 @@ A test failure looks something like this:
...
+  remove or replace such changeable text.
+  Another use of `transform` is to scrub sensitive information from the
+  snapshot.
+* `variant`: Sometimes snapshots reflect the ambient conditions, such as the
+  operating system or the version of R or one of your dependencies, and you need
+  a different snapshot for each variant. This is an experimental and somewhat
+  advanced feature, so if you can arrange things to use a single snapshot, you
+  probably should.
+  
+In typical usage, testthat will take care of managing the snapshot files below `tests/testthat/_snaps/`.
```
maelle, line 694: Ok so actually should you document the workflow for when one gets a snapshot failure and needs to fix it? (earlier when explaining snapshot failures)

- I sometimes delete the new one
- I sometimes accept it and reverse the change from the git tab

Testing.Rmd
```diff
@@ -207,210 +310,434 @@ A test failure looks something like this:
...
+
+    ```{r, error = TRUE}
+    model <- lm(mpg ~ wt, data = mtcars)
+    expect_s3_class(model, "lm")
+    expect_s3_class(model, "glm")
+    ```
+
+`expect_true()` and `expect_false()` are useful catchalls if none of the other expectations does what you need.
+
+## What to test
```
maelle, line 740: below there is some "I" in an item

maelle: reg writing test when there is a bug, maybe add a footnote about reprex, not really on topic but I was thinking about the bliss when an user gives a reprex, you put it in a test, you fix the code then close the issue

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
+To a certain extent, this is natural, since your tests need to exercise the functions in your package by doing real tasks.
+The code *inside each `test_that()` call* should be very procedural.
+But at the next level of organisation, i.e. a file with multiple tests or a folder with many test files, the step-by-step script mentality can lead you astray.
+
+Recall this advice found in section \@ref(code-r-landscape), which covers your package's "main code", i.e. everything below `R/`:
+
+> The `.R` files below `R/` should consist almost entirely of function definitions.
+> Any other top-level code is suspicious and should be carefully reviewed for possible conversion into a function.
+
+We have analogous advice for your "testing code", i.e. everything in `tests/testthat/test-*.R` files:
```
maelle, line 811: are there linters for this :thinking:

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
 ```
 
-### Building your own testing tools
+Keep reading for various alternatives that are better than having top-level code in your test files other than calls to `test_that()`.
 
-As you start to write more tests, you might notice duplication in your code.
-For example, the following code shows one test of the `floor_date()` function from `library(lubridate)`.
-There are seven expectations that check the results of rounding a date down to the nearest second, minute, hour, etc.
-There's a lot of duplication (which increases the chance of bugs), so we might want to extract common behaviour into a new function.
+### Test hygiene
```
maelle, line 835: One question here! Also related to refactoring code.

I somewhat often use the pattern

```r
is_bla <- function() {
 if (nzchar(Sys.getenv("PACKAGE_BLA_MOCK")) {
    return(FALSE)
}
  # actual code checking bla
}

```

to have a switch for testing for things such as installed packages, internet connection. Is this... good? bad?

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
+> From the book Software Engineering at Google, [Chapter 11](https://abseil.io/resources/swe-book/html/ch11.html)
+
+*Most of us don't work on a code base the size of Google. But even in a team of one, tests that you wrote six months ago might as well have been written by someone else. Especially when they suddenly start failing.*
+
+It's hard to beat the pure simplicity and obviousness of a test that creates the conditions it needs, then puts things back as they were.
+You've already seen an example of this, when we explored snapshot tests:
+
+```{r eval = FALSE}
+test_that("side-by-side diffs work", {
+  withr::local_options(width = 20)             # <-- (°_°) look here!
```
maelle, line 894: does this sound ok with a screenreader?

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
+> In its purest form, automating testing consists of three activities: writing tests, running tests, and **reacting to test failures**.
+> 
+> From the book Software Engineering at Google, [Chapter 11](https://abseil.io/resources/swe-book/html/ch11.html)
+
+Top-level code, outside of any test, is in a sort of No Man's Land:
+it's fine when your tests are all passing, but when something goes wrong and a test fails, now you have to endure extra self-inflicted pain when you least need it.
+This is the sort of practice that feels convenient when *writing tests*, but proves to be extremely inconvenient when you are *failing tests*.
+Below we recommend various robust and structured solutions to this problem.
+
+But first, seriously consider inlining the creation of a `useful_thing` whenever you need it.
```
maelle, line 1017: why not mention here the idea to make its creation a function defined centrally, if creating it takes too much space?

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
+* Attaches testthat, i.e. does `library(testthat)`.
+* Runs test helper files, i.e. executes `test/testthat/helper.R`.
+
+This eliminates the need for any `library()` calls below `tests/testthat/`, for the vast majority of R packages.
+Any instance of `library(testthat)` is clearly no longer necessary.
+Likewise, any instance of attaching one of your dependencies via `library(somePkg)` is unnecessary.
+In your tests, if you need to call functions from somePkg, do it just as you do below `R/`.
+If you have imported the function into your namespace, use `fun()`.
+If you have not, use `somePkg::fun()`.
+It's fair to say that `library(somePkg)` in the tests should be about as rare as taking a dependency via `Depends`, i.e. there is usually a better alternative.
```
maelle, line 1096: link to metadata chapter?

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
+
+Test toubleshooting nirvana looks like this:
+In a fresh R sesion, you can do `devtools::load_all()` and immediately run an individual test or walk through it line-by-line.
+There is no need to hunt around for code that has to be run first, that is found elsewhere in the test file or in a different file altogether (i.e. some file that is not executed by `load_all()`).
+
+One other function that basically should never appear below `tests/testhat/` is `source()`.
+There are several special files with an official role in testthat workflows (see below), not to mention the entire R package machinery, that provide better ways to make functions, objects, and other logic available in your tests.
+
+## Files relevant to testing  {#tests-files-overview}
```
maelle, line 1105: unsurprisingly I am thinking about my blog post table which is my mental model :grin:

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
-  expect_floor_equal("hour",   "2009-08-03 12:00:00")
-  expect_floor_equal("day",    "2009-08-03 00:00:00")
-  expect_floor_equal("week",   "2009-08-02 00:00:00")
-  expect_floor_equal("month",  "2009-08-01 00:00:00")
-  expect_floor_equal("year",   "2009-01-01 00:00:00")
+#### Built-in `skip()` functions
+
+Similar to testthat's built-in expectations, there is a family of `skip()` functions that anticipate some common situations.
+These functions often relieve you of the need to write a custom skipper.
+Here are some examples of the most useful `skip()` functions:
```
maelle, line 1649: state that the offline one also skips on CRAN from a certain testthat version?

Testing.Rmd
```diff
@@ -442,159 +769,1106 @@ It's hard to give good general advice about writing tests, but you might find th
...
+
+Our main advice about mocking is to avoid it if you can.
+This is not an indictment of mocking, but just a realistic assessment that mocking introduces new complexity that is not always justified by the payoffs.
+
+Since most R packages do not need full-fledged mocking, we do not cover it here.
+Instead we'll point you to the packages that represent the state-of-the-art for mocking in R today:
+
+* mockery: <https://github.com/r-lib/mockery>
+* mockr: <https://krlmlr.github.io/mockr/>
+* httptest: <https://enpiar.com/r/httptest/>
```
maelle, line 1697: more plugging! demo of httptest https://books.ropensci.org/http-testing/httptest.html

you should add httptest2 as well, and maybe webfakes even if it is not mocking stricto sensu

maelle (commented):
so excited for this chapter!

Testing.Rmd
```diff
@@ -207,210 +310,434 @@ A test failure looks something like this:
...
-The idea is that when a test fails, you'll know what's wrong and where in your code to look for the problem.
-You create a new test using `test_that()`, with test name and code block as arguments.
-The test name should complete the sentence "Test that ...".
-The code block should be a collection of expectations.
+We'll illustrate snapshot tests using the waldo package.
+Under the hood, testthat 3e uses waldo to do the heavy lifting of "actual vs. expected" comparisons, so it's good for you to know a bit about waldo anyway.
+One of waldo's main design goals is to present differences in a clear and actionable manner, as opposed to a frustrating declaration that "this differs from that and I know exactly how, but I won't tell you".
+Therefore, the formatting of output from `waldo::compare()` is very intentional and is well-suited to a snapshot test.
+The binary outcome of `TRUE` (actual == expected) vs. `FALSE` (actual != expected) is fairly easy to check and could get its own test.
+Here we're concerned with writing a test to ensure that differences are reported to the user in the intended way.
```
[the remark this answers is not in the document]
maelle: I'd have expected (:wink: ) an example maybe of printing the temperature hot/cold with emojis, to build upon the package of "the package within"

Testing.Rmd
```diff
@@ -442,159 +767,1118 @@ It's hard to give good general advice about writing tests, but you might find th
...
+This is not an indictment of mocking, but just a realistic assessment that mocking introduces new complexity that is not always justified by the payoffs.
+
+Since most R packages do not need full-fledged mocking, we do not cover it here.
+Instead we'll point you to the packages that represent the state-of-the-art for mocking in R today:
+
+* mockery: <https://github.com/r-lib/mockery>
+* mockr: <https://krlmlr.github.io/mockr/>
+* httptest: <https://enpiar.com/r/httptest/>
+* httptest2: <https://enpiar.com/httptest2/>
+* webfakes: <https://webfakes.r-lib.org>
```
maelle, line 1707:
```suggestion
* webfakes: <https://webfakes.r-lib.org>
* dittodb: <https://docs.ropensci.org/dittodb/>
```

jennybc:
Thanks @maelle for all of the feedback. I've taken the bits that fit in easily. But as I said elsewhere, I was close to finishing off this PR based on my own internal TODO list and I also really need to move on to other chapters.

We can continue to consider your comments later, as the chapter gets refined. For example, it probably has to be split into multiple chapters before publication.

PR #782, opened 2022-06-03
`test_file()` was deprecated, use `test_active_file()` instead

PrzeChoj:
R warned me that:

> `test_file()` was deprecated in devtools 2.4.0.
Please use `test_active_file()` instead.

So I propose this change.


hadley:
Thanks!

PR #814, opened 2022-06-28
tweak rstudio + git wording

malcolmbarrett:
this sort of sounds like `use_git()` won't work in a non-package project


jennybc:
I put the tweak in a different place because it feels weird to insert a usethis fact into an RStudio callout. But point taken.

PR #815, opened 2022-06-28
refer to @export as tag

malcolmbarrett:
instead of a directive, because it's confusing to call it that for the tag and the namespace call and more consistent with roxygen terminology


hadley:
Thanks!

PR #821, opened 2022-07-13
Grammar fix (16.2.1)

TroyVan:
You can't "run press" a key combo and a command; you press a key combo and run a command.


jennybc:
Thanks!

jennybc:
Apparently Hadley had already made an adjustment here in the meantime, so closing this.

PR #825, opened 2022-08-04
Typo and wrong body of a function

PrzeChoj:
1. On the file listing there are 2 files: `useful_thing1.rds` and `useful_thing2.rds`, but the code wrongly referenced to `useful_thing.rds`.
2. Function `is_testing()` is different from `testing_package()`.


jennybc:
Thanks! Correct on both points, but I just merged a different PR that fixes these.

PR #849, opened 2022-08-31
Mention RSPM

### LINKED ISSUE #837: Chapter 4 feedback ("Package structure and state")
I see other reviewers are leaving feedback in chunks in issues, so I'll switch to that now to be a bit more consistent.

---

I think the lead is buried in this paragraph: https://github.com/hadley/r-pkgs/blob/37aad0092f491a49800cb15aa6e7ff837621cb6f/Structure.Rmd#L117-L121

I think putting the best practice (`use_build_ignore()`) upfront is better, *then* say what it does and how to do it yourself

---

Re: a typical build ignore file: https://github.com/hadley/r-pkgs/blob/37aad0092f491a49800cb15aa6e7ff837621cb6f/Structure.Rmd#L135-L147 

I think it's worth noting that in all of these cases, using the corresponding usethis function will `.Rbuildignore` on your behalf. It seems tempting (if low-risk) to copy and paste this into the file or something

---

Worth mentioning RSPM? I've leaned on that heavily in corporate server environments, including for internal packages: https://github.com/hadley/r-pkgs/blob/37aad0092f491a49800cb15aa6e7ff837621cb6f/Structure.Rmd#L157

---

I think these two paragraphs need to be switched. Also a good chance to mention RSPM 

https://github.com/hadley/r-pkgs/blob/37aad0092f491a49800cb15aa6e7ff837621cb6f/Structure.Rmd#L163-L168

---

Related to #587, this figure might need updating to include pak:

https://github.com/hadley/r-pkgs/blob/37aad0092f491a49800cb15aa6e7ff837621cb6f/Structure.Rmd#L213

jennybc:
Closes #837

where @malcolmbarrett says:

> Worth mentioning RSPM? I've leaned on that heavily in corporate server environments, including for internal packages

### DIFF
--- a/Structure.Rmd
+++ b/Structure.Rmd
@@ -157,7 +157,7 @@ For example, `use_read_rmd()` adds "\^README\\.Rmd\$" to `.Rbuildignore`.
 If you want to distribute your package to an R user who doesn't have package development tools, you'll need to provide a **binary** package.
 Like a package bundle, a binary package is a single file.
 Unlike a bundled package, a binary package is platform specific and there are two basic flavors: Windows and macOS.
-(Linux users are generally required to have the tools necessary to install from `.tar.gz` files.)
+(Linux users are generally required to have the tools necessary to install from `.tar.gz` files, although the emergence of resources like [RStudio Public Package Manager](https://packagemanager.rstudio.com) is giving Linux users the same access to binary packages as their colleagues on Windows and macOS.)
 
 Binary packages for macOS are stored as `.tgz`, whereas Windows binary packages end in `.zip`.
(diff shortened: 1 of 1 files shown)


jennybc:
@Hadley interested in your opinion re: whether to include and, if so, wording

jennybc:
RSPM could potentially be mentioned later in this section as well. Basically, in the RSPM era, should we rewrite this to treat Windows, macOS, and Linux in a more parallel fashion as opposed to (Windows, macOS) vs. Linux?

hadley:
Yeah, I think it's a good idea to include, and wording looks fine to me.

Issue #485, opened 2018-06-05
Why you don't want Depends


jennybc:
Distill these recent discussions into improved NAMESPACE advice in the book:

https://github.com/leeper/Depends

https://twitter.com/carroll_jono/status/1003905229337706498

Nice writeup @leeper!

cc @hadley

PR #865, opened 2022-10-18
Remove "ggplot2 should `Depend` on scales"

dpprdan:
I stumbled over this sentence and I think it should go. 

1. ggplot2 does not depend on scales anymore (if it ever did)
2. from my reading of #485 and [leeper/Depends](https://github.com/leeper/Depends) the sentence is not what you'd advice now (I assume)
3. avoid others tripping over this until this section is up for editing


jennybc:
Thanks for this. It got subsumed by a much larger refactoring of this entire book part in #893.

PR #871, opened 2022-10-27
More work on `man.Rmd`

jennybc:
I’ve done everything I wanted to here now.

* [x] ~Should we mention the nonstandard-but-not-entirely exotic situation of documenting `NULL` to create a topic when there's no obvious function or dataset to document?~ I've decided "no", until I come up with a good reason.
* [x] Should we have a more explicit paragraph ~or table~ on: "the order in which stuff appears in rendered help is determined by R, not by the order in which you write your roxygen comment"? ~Should we lay out the standard order in which help sections appear?~
* [x] Make sure we're linking to all relevant roxygen2 vignettes where it makes sense
* [x] Mention `@noRd` for a non-exported function that you document for yourself?
* [x] ~Mention `@keywords internal` for an exported function that you don't want to publicize too much?~ I'm letting this go.

### DIFF
--- a/man.Rmd
+++ b/man.Rmd
@@ -327,9 +327,7 @@ Here's a two-paragraph `@description` from `stringr::str_view()`:
 #' possible, matches and unusual whitespace are coloured blue and `NA`s red.
 ```
 
-Here's another example from `stringr::str_like()`, which has a bullet list in `@description`:[^man-9]
-
-[^man-9]: The markdown parser used by roxygen2 does not actually require an empty line before lists, but it's allowed and some people prefer that style.
+Here's another example from `stringr::str_like()`, which has a bullet list in `@description`:
 
 ```{r}
 #' Detect a pattern in the same way as `SQL`'s `LIKE` operator
@@ -354,10 +352,10 @@ Once you've re-derived what the function does, you'll be able to write a better 
 
 The `@details` are just any additional details or explanation that you think your function needs.
 Most functions don't need details, but some functions need a lot.
-If you have a lot of information to convey, it's a good idea to use informative markdown headings to break the details up into manageable sections[^man-10].
+If you have a lot of information to convey, it's a good idea to use informative markdown headings to break the details up into manageable sections[^man-9].
 Here's an example from `dplyr::mutate()`. We've elided some of the details to keep this example short, but you should still get a sense of how we used headings to break up the content in to skimmable chunks:
 
-[^man-10]: In older code, you might see the use of `@section title:` which was used to create sections before roxygen2 had full markdown support.
+[^man-9]: In older code, you might see the use of `@section title:` which was used to create sections before roxygen2 had full markdown support.
     If you've used these in the past, you can now turn them into markdown headings.
 
 ```{r}
@@ -515,11 +513,11 @@ Beware of spurious diffs introduced by contributors who run `devtools::document(
 ## Return value
 
 A function's output is as important as its inputs.
-Documenting the output is the job of the `@returns`[^man-11] tag.
+Documenting the output is the job of the `@returns`[^man-10] tag.
 Here the priority is to describe the overall "shape" of the output, i.e. what sort of object it is, and its dimensions (if that makes sense).
 For example, if your function returns a vector you might describe its type and length, or if your function returns a data frame you might describe the names and types of the columns and the expected number of rows.
 
-[^man-11]: For historical reasons, you can also use `@return`, but we now favor `@returns` because it reads more naturally.
+[^man-10]: For historical reasons, you can also use `@return`, but we now favor `@returns` because it reads more naturally.
 
 The `@returns` documentation for functions in stringr is straightforward because almost all functions return some type of vector with the same length as one of the inputs.
 For example, here's how `stringr::str_like()` describes its output:
@@ -528,10 +526,10 @@ For example, here's how `stringr::str_like()` describes its output:
 #' @returns A logical vector the same length as `string`.
 ```
 
-A more complicated case is the joint documentation for `stringr::str_locate()` and `str_locate_all()`[^man-12].
+A more complicated case is the joint documentation for `stringr::str_locate()` and `str_locate_all()`[^man-11].
 `str_locate()` returns an integer matrix, and `str_locate_all()` returns a list of matrices, so the text needs to describe what determines the rows and columns.
 
-[^man-12]: We'll come back how to document multiple functions in one topic in @sec-man-multiple-functions.
+[^man-11]: We'll come back how to document multiple functions in one topic in @sec-man-multiple-functions.
 
 ```{r}
 #' @returns
@@ -707,7 +705,7 @@ There are two basic options:
     #' try(bind_cols(tibble(x = 1:3), tibble(y = 1:2)))
     ```
 
--   You can wrap the code in `\dontrun{}`[^man-13], so it is never run by `example()`. The example above would look like this if you used `\dontrun{}` instead of `try()`.
+-   You can wrap the code in `\dontrun{}`[^man-12], so it is never run by `example()`. The example above would look like this if you used `\dontrun{}` instead of `try()`.
 
     ```{r}
     #' # Row sizes must be compatible when column-binding
@@ -716,7 +714,7 @@ There are two basic options:
     #' }
     ```
 
-[^man-13]: You used to be able to use `\donttest{}` for a similar purpose, but we no longer recommend it because CRAN sets a special flag that causes the code to be executed anyway.
+[^man-12]: You used to be able to use `\donttest{}` for a similar purpose, but we no longer recommend it because CRAN sets a special flag that causes the code to be executed anyway.
 
 We generally recommend using `try()` so that the reader can see an example of the error in action.
 
@@ -753,10 +751,10 @@ In the past, we recommended only using code from suggested packages inside a blo
 
 We no longer believe that approach is a good idea, because:
 
--   Our policy is to expect that suggested packages are installed when running `R CMD check`[^man-14] and this informs what we do in examples, tests, and vignettes.
+-   Our policy is to expect that suggested packages are installed when running `R CMD check`[^man-13] and this informs what we do in examples, tests, and vignettes.
 -   The cost of putting example code inside `{ ... }` is high: you can no longer see intermediate results, such as when the examples are rendered in the package's website. The cost of a package not being installed is low: users can usually recognize the associated error and resolve it themselves, i.e. by installing the missing package.
 
-[^man-14]: This is certainly true for CRAN and is true in most other automated checking scenarios, such as our GitHub Actions workflows.
+[^man-13]: This is certainly true for CRAN and is true in most other automated checking scenarios, such as our GitHub Actions workflows.
 
 In other cases, your example code may depend on something other than a package.
 For example, if your examples talk to a web API, you probably only want to run them for an authenticated user, and never want such code to run on CRAN.
--- a/man.Rmd
+++ b/man.Rmd
@@ -3,7 +3,6 @@
 ```{r, echo = FALSE}
 source("common.R")
 status("polishing")
-library(stringr) # for links
 ```
 
 ## Introduction
@@ -174,6 +173,12 @@ A block can contain text before the first tag which is called the **introduction
 [^man-6]: The name of the file is automatically derived from the object you're documenting.
 
 Throughout this chapter we'll show you roxygen2 comments from real tidyverse packages, focusing on [stringr](https://stringr.tidyverse.org), since the functions there tend to be fairly straightforward, leading to documentation that's understandable with relatively little context.
+We attach stringr here so that its functions are hyperlinked in the rendered book (more on that in section @sec-man-key-md-features).
+
+```{r}
+library(stringr)
+```
+
 Here's a simple first example: the documentation for `str_unique()`.
 
 ```{r}
@@ -220,7 +225,7 @@ We start with the introduction, which provides the title, description, and detai
 Then we cover the inputs (the function arguments), outputs (the return value), and examples.
 Next we discuss links and cross-references, then finish off with techniques for sharing documentation between topics.
 
-### Key markdown features
+### Key markdown features {#sec-man-key-md-features}
 
 For the most part, general markdown and RMarkdown knowledge suffice for taking advantage of markdown in roxygen2.
 But there are a few pieces of syntax that are so important we want to highlight them here.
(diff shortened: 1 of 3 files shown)


man.Rmd
```diff
@@ -8,34 +8,61 @@ library(stringr) # for links
...
-Instead, we include specially formatted comments *above* the source code for each function.
-Then we use the roxygen2 package to generate the `.Rd` files from these special "roxygen comments".
-There are a few advantages to using roxygen2:
+Instead, we include specially formatted "roxygen comments" above the source code for each function[^man-1].
+Then we use the roxygen2 package to generate the `.Rd` files from these special comments[^man-2]
+. There are a few advantages to using roxygen2
+:
+
+[^man-1]: The name "roxygen" is a nod to the Doxygen documentation generator, which inspired the development of an R package named roxygen.
+    Then that original concept was rebooted as roxygen2, similar to ggplot2.
```
jennybc, line 22: I really like to know where names comes from.

man.Rmd
```diff
@@ -8,34 +8,61 @@ library(stringr) # for links
...
 -   roxygen2 provides a number of tools for sharing content across documentation topics and even between topics and vignettes.
 
-In this chapter we'll focus on documenting functions, but the same ideas apply to documenting datasets, classes and generics, and packages.
-You can learn more about those important topics in `vignette("rd-other", package = "roxygen2")`.
+In this chapter we'll focus on documenting functions, but the same ideas apply to documenting datasets (@sec-documenting-data), classes and generics, and packages.
+You can learn more about those important topics in `vignette("rd-other", package = "roxygen2")`[^man-3].
+
+[^man-3]: The text `vignette("rd-other", package = "roxygen2")` is actually a great example of autolinking in the devtools ecosystem.
+    This is literally the R code you would execute in the console to navigate to a specific vignette within the roxygen2 package.
+    But in many rendered contexts, it automatically becomes a hyperlink to that same vignette in roxygen2's pkgdown website.
```
jennybc, line 41: I really want to explain this `vignette("name", package = "package")` syntax. But perhaps I can move this footnote to the to-be-writte section I want to add for "[R] Markdown + roxygen2 Greatest Hits".

man.Rmd
```diff
@@ -8,34 +8,61 @@ library(stringr) # for links
...
 
 To get started, we'll work through the basic roxygen2 workflow and discuss the overall structure of roxygen2 comments which are organised into blocks and tags.
 
-### The documentation workflow {#man-workflow}
+### The documentation workflow {#sec-man-workflow}
+
+Unlike with testthat, there's no obvious opening move to declare that you're going to use roxygen2 for documentation.
+That's because the use of roxygen2 is purely a matter of your development workflow.
+It has no effect on, e.g., the package checking or building machinery.
+We think the roxygen approach is the best way to generate your `.Rd` files, but officially R only cares about the files themselves, not how they came to be.
```
jennybc, line 52: I feel like it's helpful to really spell this out.

man.Rmd
```diff
@@ -106,31 +145,35 @@ To summarize, there are four steps in the basic roxygen2 workflow:
...
 
-[^man-2]: The name of the file is automatically derived from the object you're documenting.
+[^man-7]: The name of the file is automatically derived from the object you're documenting.
 
 Throughout this chapter we'll show you roxygen2 comments from real tidyverse packages, focusing on stringr, since the functions there tend to be fairly straightforward, leading to documentation that's understandable with relatively little context.
-Here's a simple first example: the documentation for `stringr::str_unique()`.
+Here's a simple first example: the documentation for `stringr::str_unique()`[^man-8].
+
+[^man-8]: This may not perfectly reflect the current documentation for `str_unique()`.
+    Among other reasons, we're suppressing the use of roxygen features that are introduced later in this chapter.
```
jennybc, line 164: This function actually uses `@inheritParams` but I pretend like it doesn't at this point.

man.Rmd
```diff
@@ -147,27 +190,49 @@ str_unique <- function(string, ...) {
 Here the introduction includes the title ("Remove duplicated strings") and a basic description of what the function does.
 The introduction is followed by five tags: two `@params`, one `@returns`, one `@seealso`, one `@examples`, and one `@export`.
 
-Note that each line of the block is 80 characters wide (matching the line length of the surrounding R code) and the second and subsequent lines of the long `@param` tag are indented, which makes the entire block easier to scan.
+Note that the block has an intentional line length (typically the same as that used for the surrounding R code) and the second and subsequent lines of the long `@param` tag are indented, which makes the entire block easier to scan.
```
jennybc, line 193: I don't want to promise a super-specific line length; better to express a general principle. The stringr examples aren't consistently line-wrapped at this point and we might actually need to fiddle with line length here in the book, because we're getting horizontal scrolling.

man.Rmd
```diff
@@ -147,27 +190,49 @@ str_unique <- function(string, ...) {
...
+:::
+
 The following sections go into more depth for the most important tags.
 We start with the introduction, which provides the title, description, and details.
 Then we cover the inputs (the function arguments), outputs (the return value), and examples.
 Next we discuss links and cross-references, then finish off with techniques for sharing documentation between topics.
 
+### Key markdown features
+
+TODO: call out some useful markdown+roxygen features here.
```
jennybc, line 209: I haven't written this yet, but I think it would be useful.

man.Rmd
```diff
@@ -177,9 +242,20 @@ When writing the title, it's useful to think about how it will appear in the ref
 When a user skims the index, how will they know which functions will solve their current problem?
 This requires thinking about what your functions have in common (which doesn't need to be repeated in every title) and what is unique to that function (which should be highlighted in the title).
 
-As an example, take the titles of some of the key functions in dplyr[^man-3]:
+When we wrote this chapter, we found the function titles for stringr to be somewhat disappointing.
```
jennybc, line 245: I think it's better to show the underwhelming titles first, then the good ones. End on a happy note.

man.Rmd
```diff
@@ -217,10 +284,24 @@ It's a little extra work keeping it all up to date, but the extra effort is ofte
 #' `grepl(pattern, string)`.
 ```
 
-If you want to use multiple paragraphs or a bulleted list, you can use the explicit `@description` tag[^man-4].
-Here's an example from `stringr::str_like()`, which mimics the `LIKE` operator from SQL:
+If you want more than one paragraph, you must use an explicit `@description` tag to prevent the second (and subsequent) paragraphs from being turned into the `@details`.
+Here's a two-paragraph `@description` from `stringr::str_view()`:
```
jennybc, line 288: I chose `str_view()` to get a straight-up example of 2 paragraphs. We'll show `str_like()` and a bullet list _after_ this.

I think it's worth being super explicit about the risk of accidentally creating a details section, when you really just wanted a two-paragraph description.

man.Rmd
```diff
@@ -217,10 +284,24 @@ It's a little extra work keeping it all up to date, but the extra effort is ofte
...
+#' to see how a `pattern` matches.
+#'
+#' Matches are surrounded by `<>` and unusual whitespace (i.e. all whitespace
+#' apart from `" "` and `"\n"`) are surrounded by `{}` and escaped. Where
+#' possible, matches and unusual whitespace are coloured blue and `NA`s red.
+```
+
+Here's another example from `stringr::str_like()`, which has a bullet list in `@description`:[^man-10]
+
+[^man-10]: The markdown parser used by roxygen2 does not actually require an empty line before lists, but it's allowed and some people prefer that style.
```
jennybc, line 304: This is how I handled the fact that the empty line is not necessary and our examples sometimes have it and sometimes don't.

man.Rmd
```diff
@@ -235,16 +316,21 @@ Here's an example from `stringr::str_like()`, which mimics the `LIKE` operator f
...
-The "details" are just any additional details or explanation that you think your function needs.
+The `@details` are just any additional details or explanation that you think your function needs.
 Most functions don't need details, but some functions need a lot.
-If you have a lot of information to convey, I recommend using markdown headings to break the documentation up into sections.
-Here's an example from `dplyr::mutate()`.
-We've elided some of the details to keep this example short, but you should still get a sense of how we used headings to break up the content in to skimmable chunks:
+If you have a lot of information to convey, it's a good idea to use informative markdown headings to break the details up into manageable sections[^man-11].
+Here's an example from `dplyr::mutate()`. We've elided some of the details to keep this example short, but you should still get a sense of how we used headings to break up the content in to skimmable chunks:
+
+[^man-11]: In older code, you might see the use of `@section title:` which was used to create sections before roxygen2 had full markdown support.
```
jennybc, line 332: Moved this to a footnote, which feels more parallel with how we discuss other historical stuff, like `testthat::context()` and `@return`.

man.Rmd
```diff
@@ -392,30 +476,34 @@ Then the other stringr functions use `@inheritParams str_detect` to get this det
 #'   The pattern should contain at least one capturing group.
 ```
 
+Now that we've discussed default values and inheritance we can bring up one more dilemma.
+Sometimes there's tension between giving detailed information on an argument (acceptable values, default value, how the argument is used, etc.) and making the documentation amenable to reuse in other functions (which might differ in some specifics).
+This can motivate you to assess whether it's truly worth it for related functions to handle the same input in different ways or if standardization would be beneficial.
```
jennybc, line 481: I've recently dealt with exactly this and it's not the first time!

man.Rmd
```diff
@@ -460,28 +548,39 @@ For example `readr::read_csv()` mentions what happens if there are any parsing p
...
 
-## Examples {#sec-examples}
+## Examples {#sec-man-examples}
 
 Describing what a function does is great, but *showing* how it works is even better.
 That's the role of the `@examples` tag, which uses executable R code to demonstrate what a function can do.
 Unlike other parts of the documentation where we've focused mainly on what you should write, here we'll briefly give some content advice and then focus mainly on the mechanics.
-The mechanics of examples are complex because they must never error and they're run in four different situations:
+
+The main dilemma with examples is that you must jointly satisfy two requirements:
```
jennybc, line 562: I really want to make this fundamental dilemma with examples super concrete.

man.Rmd
```diff
@@ -526,22 +625,25 @@ Also, remember that examples are not tests.
...
 
-As much as possible, keep your examples self-contained.
+Your examples should be self-contained.
 For example, this means:
 
 -   If you modify `options()`, reset them at the end of the example.
 -   If you create a file, create it somewhere in `tempdir()`, and make sure to delete it at the end of the example.
 -   Don't change the working directory.
--   Don't write to the clipboard.
--   Avoid accessing websites in examples. If the website is down, your example will fail and hence `R CMD check` will error.
```
jennybc, line 538: I moved this below, to the section about errors.

man.Rmd
```diff
@@ -583,15 +688,22 @@ There are two basic options:
...
-[^man-7]: You used to be able to use `\donttest{}` for a similar purpose, but we no longer recommend it because CRAN sets a special flag that causes the code to be executed anyway.
+[^man-14]: You used to be able to use `\donttest{}` for a similar purpose, but we no longer recommend it because CRAN sets a special flag that causes the code to be executed anyway.
 
 We generally recommend using `try()` so that the reader can see an example of the error in action.
 
+::: callout-warning
+## Submitting to CRAN
+
+For the initial CRAN submission of your package, all functions must have at least one example and the example code can't all be wrapped inside `\dontrun{}`.
+If the code can only be run under specific conditions, use the techniques below to express those pre-conditions.
```
jennybc, line 699: I took above-average care with this wording, trying to really capture how to thread this needle.

man.Rmd
```diff
@@ -602,27 +714,58 @@ For example, dbplyr is a dplyr extension package so all of its examples start wi
...
 
 ```{r}
 #' @examplesIf some_condition()
 #' some_other_function()
 #' some_more_functions()
 ```
 
-For example, googledrive uses `@examplesIf` in almost every function because the examples can only work if you have an authenticated and active connection to Google Drive as judged by `googledrive::drive_has_token()`.
-For example, here's `googledrive::drive_publish()`:
+This looks almost like the snippet just above, but has several advantages:
```
jennybc, line 756: I felt like the beauty of `@examplesIf` needed to be made more obvious.

man.Rmd
```diff
@@ -92,9 +131,9 @@ Here's what the result looks like in RStudio:
 knitr::include_graphics("images/man-add.png", dpi = 220)
 ```
 
-The default help-seeking process looks inside **installed** packages, so to see the documentation during development, devtools overrides the usual help functions with modified versions that know to consult the **source** package.
-To activate these overrides, you need to run `devtools::load_all()` once.
-So if the development documentation doesn't appear, you may need to load your package first.
+The default help-seeking process looks inside **installed** packages, so to see your package's documentation during development, devtools overrides the usual help functions with modified versions that know to consult the **source** package.
+To activate these overrides, you'll need to run `devtools::load_all()` at least once.
+If it feels like your edits to the roxygen comments aren't having an effect, double check that you have actually regenerated the `.Rd` files with and that you've loaded your package.
```
hadley, line 136: Mention that you should see "Rendering development documentation"?

man.Rmd
hadley: I don't think folks are going to be comparing this to the literal files, but it doesn't hurt to be safe

man.Rmd
hadley: I think it's worth mentioning linking, but the rest is basic markdown that I don't think we need to repeat.

man.Rmd
hadley: To me this feels like too much inside baseball for even a footnote.

man.Rmd
hadley: I like it!

hadley (commented):
LGTM!

man.Rmd
jennybc: I just pushed this section before I saw this. So there's something for you to react to now.

man.Rmd
jennybc: I'm deleting the footnote. This factoid is currently in the section I just wrote on the most important markdown stuff, so we'll see what you think.

man.Rmd
jennybc: Moved

man.Rmd
jennybc: Yeah ok we can kill this footnote. We have a lot already in this chapter.

man.Rmd
jennybc: done

man.Rmd
jennybc: I've got this little section, because I think it will actually push some useful info into people's brains. Let's see if we get any +/- feedback on it and how we feel about it later.

PR #875, opened 2022-11-04
More on which content is in vs. out

jennybc:
When revising the miscellany chapter, it felt really great to just delete some things. I'd like to plan for more of that, so it was cathartic to write this paragraph. I think this will also be a useful section to link to when people inevitably ask us to add content on topics we aren't going to cover.


hadley:
I like it!

PR #882, opened 2022-11-11
State the license in a way that GitHub recognizes

jennybc:
I'm hoping this makes the repo go from "Unknown, Unknown licenses found" to something ... better? That might require changing the name of the licensing chapter (currently `license.Rmd`) which isn't helping matters.

I've inlined the whole license which seems to be pretty common and the best way to get it recognized. I kept the "Attribution-NonCommercial-NoDerivatives" part but upgraded from 3.0 to 4.0 and changed from US to International, while I was here.

This is what I'm trying to fix:

Screen Shot 2022-11-11 at 2 23 12 PM


hadley:
Looks good, but can you confirm the updated license is what we agreed with O'Reilly?

jennybc:
We got confirmation from O'Reilly that this change is good.

> The change to the license file looks correct.
>
> The license info for R Packages, 2e is
>
> Creative Commons Attribution-Noncommercial-No Derivatives 4.0 International Public License

PR #885, opened 2022-11-17
Rstudio->Posit

howardbaek:
RStudio Public Package Manager -> Posit Public Package Manager


jennybc:
Thanks! I have updated to a different URL as well.

PR #902, opened 2022-12-14
Actually write the website chapter

jennybc:
My overall goal is to not repeat too much from the pkgdown docs, but to really highlight things that I do and think about regularly in my pkgdown experience.


website.Rmd
```diff
@@ -2,17 +2,189 @@
...
+
+-   The convention is to orient the logo with a vertex at the top and bottom, with flat vertical sides.
+
+-   If you think you might print stickers, make sure to comply with the de facto standard for sticker size.
+    [hexb.in](http://hexb.in/sticker.html) is a reliable source for the dimensions and also provides a list of potential vendors for printed stickers.
+
+-   The [hexSticker package](https://cran.r-project.org/package=hexSticker) helps you make your logo from within the comfort of R.
+
+Once you have your logo, the `usethis::use_logo()` function places an appropriately scaled copy of the image file at `man/figures/logo.png` and also provides a copy-paste-able markdown snippet to include your logo in your `README`.
+pkgdown will also discover a logo placed in the standard location and incorporate it into your site.
```
jennybc, line 182: Is this true? :thinking:

website.Rmd
```diff
@@ -2,17 +2,189 @@
...
+-   Vignettes (and articles) (@sec-vignettes).
+
+-   README and NEWS (@sec-other-markdown).
+
+Wouldn't it be divine if all of that somehow got bundled up together into a beautiful website for your package?
+The [pkgdown package](https://pkgdown.r-lib.org) is meant to provide exactly this magic and that is the topic of this chapter.
+
+## Initiate a site
+
+Assuming your package has a valid structure, pkgdown should be able to make a website for it!
```
hadley, line 25: I think somewhere we need to reinforce that the default display is pretty good, so you can be up and running with a decent website in ~5 minutes at a minimum. I think a lot of people putting off making a website because they think it's going to be a lot of work.

website.Rmd
```diff
@@ -2,17 +2,189 @@
...
+#| eval: false
+pkgdown::build_site()
+```
+
+```{r}
+#| echo: false
+usethis::with_project(temp_pkg, code = pkgdown::build_site(), setwd = TRUE)
+```
+
+You'll typically be in an interactive session when you do this, in which case your newly rendered site should appear in your default web browser.
```
hadley, line 79: I think you can remove the repeated "typically" bit

website.Rmd
```diff
@@ -2,17 +2,189 @@
...
+It's fun to have a package logo!
+In the R community, we have a strong tradition of hex stickers, so it can be nice to join in with a hex logo of your own.
+Keen R user Amelia McNamara [made herself a dress](https://www.amelia.mn/blog/misc/2019/08/17/Tidy-Dress.html) out of custom hex logo fabric and useR!
+2018 featured a [spectacular hex photo wall](https://www.mitchelloharawild.com/blog/hexwall/).
+
+Here are some resources to guide your logo efforts:
+
+-   The convention is to orient the logo with a vertex at the top and bottom, with flat vertical sides.
+
+-   If you think you might print stickers, make sure to comply with the de facto standard for sticker size.
```
hadley, line 176: I think you should inline the size advice here (at least the aspect ratio)

website.Rmd
hadley: Yes — it now automatically adds to most pages (e.g. https://ggplot2.tidyverse.org/reference/index.html)

website.Rmd
jonthegeek: I very strongly agree with this sentiment! I kept putting off "learning pkgdown" and then it turned out it was absolutely painless.

jennybc: ok added a tip about this

website.Rmd
jennybc: done

website.Rmd
jennybc: added the official "spec" as an image

website.Rmd
```diff
@@ -2,17 +2,297 @@
...
+
+    ``` yaml
+    URL: https://dplyr.tidyverse.org, https://github.com/tidyverse/dplyr
+    ```
+
+-   Your `_pkgdown.yml` file must include the URL for your site:
+
+    ``` yaml
+    url: https://dplyr.tidyverse.org
+    ```
```
jennybc, line 232: @Hadley I'm looking for some confirmation on this. In particular, I feel like I've read somewhere that the pkgdown site's URL must be in DESCRIPTION but then it neglects to mention that GitHub must be there too (obviously I should try to remember where I saw that).

Yet it seems like the GitHub URL would also be a requirement, yeah? In order to find `_pkgdown.yml`?

jennybc:
@Hadley I absorbed yesterday's feedback and have added the "Reference index" section, in case you want to take another pass.

website.Rmd
jennybc: I must be thinking of the language from pkgdown's vignette on linking:

https://pkgdown.r-lib.org/articles/linking.html#across-packages

> In order for a pkgdown site to be findable, it needs to be listed in two places:
> 
> * In the URL field in the DESCRIPTION, as in dplyr:
> 
>     URL: https://dplyr.tidyverse.org, https://github.com/tidyverse/dplyr
> * In the url field in `_pkgdown.yml`, as in dplyr
>
>    url: https://dplyr.tidyverse.org
> ...
> Now, when you build a pkgdown site for a package that links to the dplyr documentation (e.g., `dplyr::mutate()`), pkgdown looks first in dplyr’s DESCRIPTION to find its website, then it looks for `pkgdown.yml`, and uses the metadata to generate the correct links.

Now that I've done a close reading, I conclude that, while a good idea, the GitHub URL has nothing to do with pkgdown's auto-linking. Instead this is the critical file, while comes from the pkgdown site itself:

> pkgdown generate a public facing `pkgdown.yml` file that provides metadata about the site

jennybc:
This covers everything I want it to now.

website.Rmd
```diff
@@ -2,17 +2,396 @@
...
+pkgdown creates a function reference in `reference/` that includes one page for each `.Rd` help topic in `man/`.
+This is one of the first pages you should admire in your new site.
+As you look around, there are a few things to contemplate, which we review below.
+
+pkgdown executes all your examples (@sec-man-examples) and inserts the rendered results.
+We find this is a fantastic improvement over just showing the source code.
+This view of your examples can be eye-opening and often you'll notice things you want to add, omit, or change.
+If you're not satisfied with how your examples appear, this is a good time to review techniques for including code that is expected to error (@sec-man-examples-errors) or that can only be executed under certain conditions (@sec-man-examples-dependencies-conditional-execution).
+
+These help topics will be linked to from many locations within and, potentially, beyond your pkgdown site.
```
hadley, line 209: Subsections for linking, index organisation, etc?

website.Rmd
```diff
@@ -2,17 +2,396 @@
...
+
+To learn more, see `?pkgdown::build_reference`.
+
+## Vignettes and articles
+
+@sec-vignettes deals with vignettes, which are long-form guides for a package.
+They afford various opportunities beyond what's possible in function documentation.
+For example, it's much easier to work with code that produces an error, it's less cumbersome to work with suggested packages, and there are ways to shield the reader from setup and teardown that does not apply to authentic usage.
+
+A package's vignettes appear, in rendered form, in its website, in the "Articles" dropdown menu.
```
hadley, line 299: Maybe work in something like: A vignette is technical term that we don't expect all R users to know; that's why pkgdown uses the more general articles here.

hadley (approved):

website.Rmd
jennybc: added

website.Rmd
jennybc: I've added this point.

PR #918, opened 2023-01-08
remove spacing between highlighting comments and code

zkamvar:
on narrow screens, the `# <-- look here` style inline comments are often hidden in the overflow.


jennybc:
There's a new code annotation feature coming to Quarto, so we might be able to use that instead, in the long run. And I'm not sure that "small screens" are a high priority for the book (or, at least, not a higher priority than "large screens"). But I also don't care deeply about this, so am happy to merge.

PR #923, opened 2023-01-11
Revise chapter on lifecycle

### LINKED ISSUE #482: Advice on making a package with multiple versions of another package
Like with the development and CRAN version so that you can release unconditionally (without waiting for the other package)

* `packageVersion("ggplot2") >= "2.3.0"` won't work until the package is actually released
* `packageVersion("ggplot2") > "2.2.1"` will fail if we have to do a hot-patch release
* `packageVersion("ggplot2") >= "2.2.99"`

jennybc:
Closes #482

### DIFF
--- a/lifecycle.Rmd
+++ b/lifecycle.Rmd
@@ -28,6 +28,65 @@ We can then connect transitions between these stages to concrete and specific ac
 
 Finally, in ??
 we delve into the specifics of releasing a package on CRAN and the ongoing work required to keep it safely on CRAN.
+We've been calling out CRAN-specific concerns throughout the book.
+So ??
+is where we underscore some of the most crucial points and, most importantly, can take a holistic look at blah blah CRAN.
+
+## Package evolution
+
+First we should establish a working definition of what it means for your package to change.
+Technically, you could say that the package has changed every time any file in its source changes.
+This level of pedantry isn't terribly useful, though.
+The smallest increment of change that's meaningful is probably a Git commit.
+This represents a specific state of the source package that can be talked about, installed from, compared to, subjected to `R CMD check`, reverted to, and so on.
+This level of granularity is really only of interest to developers.
+But the package states accessible via the Git history are genuinely useful for the maintainer, so if you needed any encouragement to be more intentional with your commits, let this be it.
+
+The primary user-facing signal of change is the package version number.
+Recall that this important piece of metadata lives in `Version` field of the `DESCRIPTION` file:
+
+``` yaml
+Package: usethis
+Title: Automate Package and Project Setup
+Version: 2.1.6
+...
+```
+
+If you visit the CRAN landing page for usethis, you can access its history via *Downloads \> Old sources \> usethis archive*.
+That links to a folder of package bundles, reflecting usethis's source for each version released on CRAN:
+
+```{r}
+#| echo: false
+library(tidyverse)
+library(pkgsearch)
+cran_package_history("usethis") |> 
+  select(Version, `Date/Publication`)
+```
+
+This is the type of package evolution we're going to address in this chapter.
+In section ??,
+we'll delve into the world of software version numbers, which is a richer topic than you might expect.
+Then we'll go over specific rules and tools around version numbers for R packages.
+Finally, we'll explain the conventions we use for the version numbers of tidyverse packages.
+
+But first, this is a good time to revisit a resource we first pointed out in ??,
+when introducing the different states of an R package.
+Recall that the (unofficial) `cran` organization on GitHub provides a read-only history of all CRAN packages.
+For example, you can get a different view of usethis's released versions at <https://github.com/cran/usethis/>.
+
+The archive provided by CRAN itself allows you to download older versions of usethis as `.tar.gz` files, which is useful if you truly want to get your hands on the source of an older version.
+However, if you just want to quickly check something about a version or compare two versions of usethis, the read-only GitHub mirror is much more useful.
+Each commit in this repo's history represents a CRAN release, which makes it easy to see exactly what changed: <https://github.com/cran/usethis/commits/HEAD>.
+
+This information is technically available from the repository where usethis is actually developed (<https://github.com/r-lib/usethis>).
+But you would have to work much harder to zoom out to the level of CRAN releases, as opposed to the small incremental steps in which development actually unfolds.
+These three different views of usethis's evolution are all useful for different purposes:
+
+-   <https://cran.r-project.org/src/contrib/Archive/usethis/>
+
+-   <https://github.com/cran/usethis/commits/HEAD>
+
+-   <https://github.com/r-lib/usethis/commits/HEAD>
 
 ## Version
 
--- a/lifecycle.Rmd
+++ b/lifecycle.Rmd
@@ -22,27 +22,17 @@ Even if your priorities differ from those of the tidyverse team, this chapter sh
 
 Very few users complain when a package gains features or gets a bug fix.
 Instead, we're mostly going to talk about so-called breaking changes, such as removing a function or narrowing the acceptable inputs for a function.
-
 In @sec-lifecycle-breaking-change-definition, we explore how to determine whether something is a breaking change or, more realistically, to gauge where it lies on a spectrum of "breakingness".
 Even though it can be painful, sometimes a breaking change is beneficial for the long-term health of a package (@sec-lifecycle-breaking-change-pros-cons).
 
 Since change is inevitable, the kindest thing you can do for your users to communicate clearly and help them adapt to change.
 Several practices work together to achieve this:
 
--   Package version number: be intentional about what sort of changes are appropriate in, e.g., a patch release versus a major release (@sec-lifecycle-evolution, @sec-lifecycle-version-number).
+-   Package version number: The main form of user-facing change is a package release (@sec-lifecycle-evolution). Be intentional about what sort of changes are included in, e.g., a patch release versus a major release (@sec-lifecycle-version-number, @sec-lifecycle-release-type).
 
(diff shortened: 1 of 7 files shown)


lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+
+-   The maintainer(s), which includes you and possibly others (especially in the future).
+
+-   The existing users, which could be just you or a small group of colleagues or it could be tens of thousands of people.
+
+-   The future users, which hopefully includes the existing users, but could potentially include many more people.
+
+It's impossible to optimize for all of these folks, all of the time, all at once.
+So we'll describe how we think about various trade-offs.
+Even if your priorities differ from ours, this chapter should still help you identify issues you want to consider.
```
hadley, line 22: Reinforce ours = tidyverse?

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+The version number of your package is always increasing, but it's more than just an incrementing counter -- the way the number changes with each release can convey information about the nature of the changes.
+The transition from 0.3.1 to 0.3.2, which is a patch release, has a very different vibe from the transition from 0.3.2 to 1.0.0, which is a major release.
+A package version number can also convey information about where the package is in its lifecycle.
+For example, the version 1.0.0 often signals that the public interface of a package is considered stable.
+
+How do you decide which type of release to make, i.e. which component(s) of the version should you increment?
+A key concept is whether the associated changes are backward compatible, meaning that pre-existing code will still "work" with the new version.
+We put "work" in quotes, because this designation is open to a certain amount of interpretation.
+A hardliner might take this to mean "the code works in exactly the same way, in all contexts, for all inputs".
+A more pragmatic interpretation is that "the code still works, but could produce a different result in some edge cases".
```
hadley, line 198: I think you need the "f your package had a bug in it" content here.

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+A change that is not backward compatible is often described as a *breaking* change.
+In section @sec-lifecycle-backward-compatibility, we'll revisit the topic of backward compatibility.
+Similar to the issue of package dependencies (@sec-dependencies-pros-cons), it's a question of trade-offs.
+
+In practice, backward compatibility is not a clear-cut distinction.
```
hadley, line 203: Our definition is something like: A breaking changes causes a CRAN packages that previously passed R CMD check to now fail R CMD check AND that the previously behaviour was correct. (i.e. if we make a change to a package that reveals a bug in your package, we don't consider that to be a breaking change).

But maybe we need different terminology for those cases?

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+It is typical to assess the impact of a change from a few angles:
+
+-   Degree of change in behaviour.
+    The most extreme is to make something that used to be possible into an error, i.e. impossible.
+    In contrast, changes to a print method are much less significant.
```
hadley, line 208: But even changing a minor detail of printing can cause a package to fail it's check, if the tests were written in a particularly fragile why.

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
-`usethis::create_package()` does this, by default.
-`usethis::use_version()` increments the package version; when called interactively, with no argument, it presents a helpful menu:
+## Major vs. minor vs. patch release
+
+The version number of your package is always increasing, but it's more than just an incrementing counter -- the way the number changes with each release can convey information about the nature of the changes.
+The transition from 0.3.1 to 0.3.2, which is a patch release, has a very different vibe from the transition from 0.3.2 to 1.0.0, which is a major release.
+A package version number can also convey information about where the package is in its lifecycle.
+For example, the version 1.0.0 often signals that the public interface of a package is considered stable.
+
+How do you decide which type of release to make, i.e. which component(s) of the version should you increment?
```
hadley, line 194: Maybe it's worth distinguishing theoretical breaking vs actual breaking. Sometimes a theoretical break doesn't actually affect any code.

hadley: Maybe it's worth specifically listing some breaking changes:

* Removing a function
* Remove an argument
* Decreasing the set of valid inputs to a function

And some things that generally aren't breaking changes:

* Adding a function (although there's some small chance on introducing a conflict in user code)
* Adding an argument
* Increasing the set of valid inputs.
* Changing the text of a print method or warning/error.

And then talk about things on the boundary, like fixing bugs?

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+Recall that a version number will have one of these forms, if you're following our conventions:
+
+``` default
+<major>.<minor>.<patch>        # released version
+<major>.<minor>.<patch>.<dev>  # in-development version
+```
+
+If the current package version is `0.8.1.9000`, here's our advice on how to pick the version number for the next release:
+
+-   Increment `patch`, e.g. `0.8.2` for a **patch release**: you've fixed bugs, but you haven't added any significant new features.
```
hadley, line 225: Should not have breaking changes.

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+    This is best reserved for changes that are not backward compatible and that are likely to affect many users.
+    The `1.0.0` release has special significance and typically indicates that your package is feature complete with a stable API.
 
-```{r, eval = FALSE}
+[^lifecycle-4]: For some suitably pragmatic definition of "backward compatible".
+
+The trickiest decision you are likely to face is whether a change is "breaking" enough to deserve a major release.
+For example, if you make an API-incompatible change to a rarely-used part of your code, it may not make sense to increase the major number.
+But if you fix a bug that many people depend on (it happens!), it will feel like a breaking change to those folks.
+It's conceivable that such a bug fix could merit a major release.
```
hadley, line 243: I think it's worth mentioning something about marketing here — people are more likely to pay attention to major releases and you want to take advantage of that.

lifecycle.Rmd
```diff
@@ -7,52 +7,251 @@ status("drafting")
...
+``` default
+<major>.<minor>.<patch>        # released version
+<major>.<minor>.<patch>.<dev>  # in-development version
+```
+
+If the current package version is `0.8.1.9000`, here's our advice on how to pick the version number for the next release:
+
+-   Increment `patch`, e.g. `0.8.2` for a **patch release**: you've fixed bugs, but you haven't added any significant new features.
+    For example, if we discover a show-stopping bug shortly after a release, we would make a quick patch release with the fix.
+    Most releases will have a patch number of 0.
```
hadley, line 227: Maybe say we don't do blog posts about patch releases because we don't think people know about it?

lifecycle.Rmd
```diff
@@ -65,78 +264,193 @@ usethis::use_version()
...
+Therefore, package maintainers still need to exercise considerable caution and care when they introduce breaking changes, regardless of what's happening with the version number.
+In section ??,
+we describe how tidyverse packages approach this, supported by tools in the lifecycle package.
+
+As with dependencies (@sec-dependencies-pros-cons), we find that extremism isn't a very productive stance.
+Extreme resistance to breaking change puts a significant drag on ongoing development and maintenance.
+Backward compatible code tends to be harder to work with because of the need to maintain multiple paths to support functionality from previous versions.
+The harder you strive to maintain backward compatibility, the harder it is to develop new features or fix old mistakes.
+This, in turn, can discourage adoption by new users and can make it harder to recruit new contributors.
+Be concerned about backward compatibility, but don't let it paralyze you.
```
hadley, line 293: OTOH if you constantly make breaking changes, people will find it very frustrating to use your package and possibly stop. Find a happy medium.

lifecycle.Rmd
```diff
@@ -65,78 +264,193 @@ usethis::use_version()
...
+# old function
+plus3 <- function(x, y, z) {
+  lifecycle::deprecate_warn("1.0.0", "plus3()", "add3()")
+  add3(x, y, z)
+}
+
+plus3(1, 2, 3)
+```
+
+At this point, a user who calls `plus3()` sees a warning explaining that the function has a new name, but we go ahead and call `add3()` with their inputs.
```
hadley, line 411: Because warning every time a function is used is very annoying (almost as annoying as removing the function), by default `deprecate_warn()` will only warn once every 8 hours. You can make it warn on every use by setting `always = TRUE`.

Lifecycle is trying to walk a tightrope — it's goal is to be just annoying enough to get the user to fix the code before the function or argument goes away, but not annoying enough that it substantially hampers their quality of life.

lifecycle.Rmd
jennybc: Done (but not pushed yet)

lifecycle.Rmd
jennybc: This has been incorporated now.

lifecycle.Rmd
jennybc: This section has been heavily reworked and this now comes earlier.

lifecycle.Rmd
jennybc: I worked these specifics in.

lifecycle.Rmd
jennybc: Incorporated

lifecycle.Rmd
jennybc: Added

lifecycle.Rmd
jennybc: Added

lifecycle.Rmd
jennybc: Incorporated

lifecycle.Rmd
jennybc: Added

lifecycle.Rmd
jennybc: Incorporated

PR #927, opened 2023-01-18
Dependencies: change internal link branch name

zkamvar:
There was a link that went to the `master` branch of the actions repository. I've changed it to be `HEAD` so that it redirects to the right place.


jennybc:
Thanks!

PR #929, opened 2023-02-07
Text suggestions for "The Whole Game"

### LINKED ISSUE #790: suggestions for Chapter 2: The Whole Game
N.B. I was asked to review Chapters 1-9 by O'Reilly and instructed to leave specific comments here with the understanding of one issue per chapter, ignoring the TODOs:

I am a huge fan of the cake-first model apprach for exposing some of the aspects of package development readers will get using this book. The exposition of the workflow and demonstration of the package on GitHub is very effective. I am definitely in favour of the inventory section at the bottom as has been suggested in https://github.com/hadley/r-pkgs/issues/622#issuecomment-1051275664.

I have a few suggestions for improvement: 

### Manage Anxiety

This chapters very much stands on its own as the toy package is not used at all
for the rest of the book (there are two references to the chapter in subsequent chapters). 
With this in mind, it would help minimize anxiety for readers if it was explicitly mentioned
that there will be no further iteration on {regexcite}.

In addition, I would suggest to reiterate that readers should understand how to load a package and write a function. 

### Remind readers about using RStudio

This came up in https://github.com/hadley/r-pkgs/blob/31eb7c67788cfc14511668d96615c6ad4be6ea66/Whole-game.Rmd#L176-L184

Note for `create_package()`: This is partially addressed in 5.1.3, but if user
is not working in RStudio, they will only have DESCRIPTION, NAMESPACE, and R/.
While the chapter does state at the beginning that RStudio is being used, the
missing `.Rbuildignore` and `.gitignore` files are a bit surprising since these
are not RStudio-specific. This could be confusing for people who are not using
RStudio (despite your best efforts to nudge them in the right direction, water
and horses and all that).

### Display tables, not tibbles

https://github.com/hadley/r-pkgs/blob/31eb7c67788cfc14511668d96615c6ad4be6ea66/Whole-game.Rmd#L245-L250

The file table could be presented as a table by adding `knitr::kable()` to the
call chain to avoid extra cognitive load trying to figure out where the table
came from.

ateucher:
This is just a few suggestions in "The Whole Game" chapter, mostly for clarity and to provide a bit of context in a few places. Closes #790

### DIFF
--- a/whole-game.Rmd
+++ b/whole-game.Rmd
@@ -48,7 +48,7 @@ We strongly suggest you follow along and create this toy package with us, howeve
 
 You can initiate your new package from any active R session.
 You don't need to worry about whether you're in an existing or new project or not.
-The functions we use make sure that the package we will create is in a new, clean, project.
+The functions we use ensure that we create a new clean project for the package.
 
 Load the devtools package, which is the public face of a set of packages that support various aspects of package development.
 The most obvious of these is the usethis package, which you'll see is also being loaded.
--- a/whole-game.Rmd
+++ b/whole-game.Rmd
@@ -189,7 +189,7 @@ In the *Files* pane, go to *More (gear symbol) \> Show Hidden Files* to toggle t
 A select few are visible all the time, but sometimes you want to see them all.
 :::
 
--   `.Rbuildignore` lists files that we need to have around but that should not be included when building the R package from source. By default, this file (as well as `.gitignore`) is only created by `create_package()` if you are using RStudio, however you will likely need to make use of it in the future regardless of what editor you are using. It is discussed in more detail in @sec-rbuildignore.
+-   `.Rbuildignore` lists files that we need to have around but that should not be included when building the R package from source. If you aren't using RStudio, `create_package()` may not create this file (nor `.gitignore`) at first, since there's no RStudio-related machinery that needs to be ignored. However, you will likely develop the need for `.Rbuildignore` at some point, regardless of what editor you are using. It is discussed in more detail in @sec-rbuildignore.
 -   `.Rproj.user`, if you have it, is a directory used internally by RStudio.
 -   `.gitignore` anticipates Git usage and tells Git to ignore some standard, behind-the-scenes files created by R and RStudio. Even if you do not plan to use Git, this is harmless.
 -   `DESCRIPTION` provides metadata about your package. We edit this shortly.
--- a/whole-game.Rmd
+++ b/whole-game.Rmd
@@ -42,11 +42,13 @@ It's meant to paint the Big Picture and suggest a workflow, before we descend in
 
 To keep the pace brisk, we exploit the modern conveniences in the devtools package and the RStudio IDE. In later chapters, we are more explicit about what those helpers are doing for us.
 
+We strongly suggest you follow along and create this toy package with us, however if you just want to read through it (or something goes awry halfway through!), completing it is not critical to understanding the remainder of the book.
+
 ## Load devtools and friends
 
 You can initiate your new package from any active R session.
 You don't need to worry about whether you're in an existing or new project or not.
-The functions we use take care of this.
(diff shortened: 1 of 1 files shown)


whole-game.Rmd
```diff
@@ -42,11 +42,13 @@ It's meant to paint the Big Picture and suggest a workflow, before we descend in
 
 To keep the pace brisk, we exploit the modern conveniences in the devtools package and the RStudio IDE. In later chapters, we are more explicit about what those helpers are doing for us.
 
+We strongly suggest you follow along and create this toy package with us, however if you just want to read through it (or something goes awry halfway through!), completing it is not critical to understanding the remainder of the book.
```
jennybc, line 45: Hmmm, I think this is close but not quite right. If this chapter is impenetrable, I think the rest of the book is going to be rough sailing indeed. I think we should only make a more limited promise that this chapter is self-contained.

whole-game.Rmd
```diff
@@ -42,11 +42,13 @@ It's meant to paint the Big Picture and suggest a workflow, before we descend in
...
+
 ## Load devtools and friends
 
 You can initiate your new package from any active R session.
 You don't need to worry about whether you're in an existing or new project or not.
-The functions we use take care of this.
+The functions we use make sure that the package we will create is in a new, clean, project.
```
jennybc, line 51:
```suggestion
The functions we use ensure that we create a new clean project for the package.
```

whole-game.Rmd
```diff
@@ -187,9 +189,9 @@ In the *Files* pane, go to *More (gear symbol) \> Show Hidden Files* to toggle t
 A select few are visible all the time, but sometimes you want to see them all.
 :::
 
--   `.Rbuildignore` lists files that we need to have around but that should not be included when building the R package from source. More in @sec-rbuildignore.
+-   `.Rbuildignore` lists files that we need to have around but that should not be included when building the R package from source. By default, this file (as well as `.gitignore`) is only created by `create_package()` if you are using RStudio, however you will likely need to make use of it in the future regardless of what editor you are using. It is discussed in more detail in @sec-rbuildignore.
```
jennybc, line 192: I wasn't really conscious of this, but I did the experiment and see what you mean. Now that I get it, I'm going to take a crack at rewording.

```suggestion
-   `.Rbuildignore` lists files that we need to have around but that should not be included when building the R package from source. If you aren't using RStudio, `create_package()` may not create this file (nor `.gitignore`) at first, since there's no RStudio-related machinery that needs to be ignored. However, you will likely develop the need for `.Rbuildignore` at some point, regardless of what editor you are using. It is discussed in more detail in @sec-rbuildignore.
```

jennybc (commented):
Thanks, this is definitely improving the chapter.

whole-game.Rmd
ateucher: I like that, it's clearer

whole-game.Rmd
ateucher: Right, that's a good point!

whole-game.Rmd
jennybc: Forgive me if you did this or already know this, but just in case, you can accept suggestions like this by clicking on "Commit suggestion" in this PR interface here in the browser. Then it will give you a chance to commit right away or build up a batch of suggestions and commit as one. Then you'd want to pull these new changes into your local branchsd right away.

ateucher: Yup, I know and love that feature!

whole-game.Rmd
ateucher: Something like: "We strongly suggest you follow along and create this toy package with us, however it is self-contained and completing it is not a requirement to advancing through the remainder of the book."?

ateucher: Pushed an alternate wording that is hopefully better.

ateucher:
I believe this addresses everything in #790 so I added the "closes" keyword to the description

jennybc:
Thanks!

PR #930, opened 2023-02-07
Format tables in whole-game with kable

ateucher:
As suggested in #790. I also truncated the commit hash in the git log table since it took up a lot of horizontal space.

Screenshot attached here so you don't have to render the book to preview it:

image


jennybc:
Thanks!

Issue #851, opened 2022-09-01
Update the `.libPaths()` content


jennybc:
https://github.com/hadley/r-pkgs/blob/1e3f025b713523501381cda1c112446f8cfec912/Structure.Rmd#L271

Should show with a more recent R version (not 3.6).

Re: having a user library, consider adding coverage for linux. In #801 @jthomasmock recommended https://stackoverflow.com/a/38417179 as a resource.

jennybc:
I think this is handled up to what we're going to do (if we're doing to do anything) about linux.

PR #935, opened 2023-02-10
Structure: Update `.libPaths()` section

ateucher:
Update versions and `.libPaths()` patterns, provide code to create `R_LIBS_USER` on macOS. Partially addresses #851, but does not do anything for linux.

### DIFF
--- a/structure.Rmd
+++ b/structure.Rmd
@@ -305,8 +305,8 @@ Here's how this might look on Windows:
 ```{r, eval = FALSE}
 # on Windows
 .libPaths()
-#> [1] "C:/Users/jenny/Documents/R/win-library/3.6"
-#> [2] "C:/Program Files/R/R-3.6.0/library"
+#> [1] "C:/Users/jenny/Documents/R/win-library/4.2"
+#> [2] "C:/Program Files/R/R-4.2.2/library"
 
 lapply(.libPaths(), list.dirs, recursive = FALSE, full.names = FALSE)
 #> [[1]]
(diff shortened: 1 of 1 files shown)


structure.Rmd
```diff
@@ -330,8 +330,8 @@ Here's a similar look on macOS (but your results may vary):
 ```{r, eval = FALSE}
 # on macOS
 .libPaths()
-#> [1] "/Users/jenny/Library/R/3.6/library"
-#> [2] "/Library/Frameworks/R.framework/Versions/3.6/Resources/library"
+#> [1] "/Users/jenny/Library/R/arm64/4.2/library"
+#> [2] "/Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/library"
```
jennybc, line 334: I assume this is based on what you see on your machine. I'm still on Intel, so `.libPaths()` looks different for me now, both vs. what was originally here _and_ different from what you're proposing. But I think your version is more forward-looking.

``` r
.libPaths()
#> [1] "/Users/jenny/Library/R/x86_64/4.2/library"                     
#> [2] "/Library/Frameworks/R.framework/Versions/4.2/Resources/library"
```

<sup>Created on 2023-02-10 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

jennybc (approved):

structure.Rmd
ateucher: Yeah, the architecture is now explicitly part of the path in `R_LIBS_USER` according to [`?libPaths`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/libPaths.html)

PR #936, opened 2023-02-13
Edits to System Setup.

### LINKED ISSUE #791: suggestions for Chapter 3: System setup
N.B. I was asked to review Chapters 1-9 by O'Reilly and instructed to leave specific comments here with the understanding of one issue per chapter, ignoring the TODOs:

Overall, this chapter sets folks up for success well. I have some questions and suggestions for the deal with devtools and usethis (which originated in #563)

## Misc

https://github.com/hadley/r-pkgs/blob/31eb7c67788cfc14511668d96615c6ad4be6ea66/Setup.Rmd#L33

suggestion to replace `foo` with `pkg`.

## devtools, usethis, and you

Summed up in https://github.com/hadley/r-pkgs/pull/563#issuecomment-497838205, the point of the chapter is this:

> - Settle once and for all what our recommendations actually are!
> - Develop a clean, only-what-you-need-to-know story about the devtools breakup
> - Concrete instructions on system setup

As someone who has to deal with a tool's long and storied history, I can feel how hard this section is to write in my bones. That being said, I think there could be improvements on the first two points. 

### The weight of history

> After 7 years of development, devtools had grown into a rather unwieldy package, making maintenance difficult. 

I would suggest for the authors to avoid deprecation of the tool the readers will be using throughtout the book as it can be demotivating. It makes sense to set expectations for the reader early, but for someone who is just getting into package development, the history of why is inside baseball (though it does serve well as an example for package scope). My recommendation is to remove the deprecating language and put the history in a footnote for the intrepid and curious readers.

### Dueling narratives

When I read through the section, I was not clear about what its objectives were (until I found https://github.com/hadley/r-pkgs/pull/563#issuecomment-497838205).

As it reads, this section is split into two messages. The first message seems to say "load devtools interactively and use its functions interactively." It then mentions using the functionality from
devtools programmatically, but it's not clear _why_ someone would want to do this since much of the packages are geared for package management and development, which requires the knowlege that this book will teach.

The other message of this section seems to be: "If you have a problem with devtools, figure out what package its coming from and report it there." This can be confusing because it's not always clear where the functions are coming from (as an author of a package that uses the `Depends` field, I know this struggle all too well). 

### Maybe a solution?

The confusion comes in when you read through the narrative, it's organised like so:

````markdown
Historical context

 - useR
 - developR
   - caveat
   - caveat
 - reporting (encompasses useR and developR)

useR example
```
useR mode
```

developR example
```
developR mode
```

usethis: first useR then developR
````

Instead, the message might be more clear if the useR perspective were presented first, explaining reporting, examples, and usethis, followed by the developR perspective:

````markdown
Brief historical context

useR perspective
```
useR example
```
reporting bugs
usethis in relation to devtools (`use_` and `create_` functions)

developR perspective
caveats
```
developR example
```
````

For reporting errors and for the developR perspective, it might be worthwhile to provide an auxillary table of the seven packages that devtools uses from the great uncoupling and what devtools functions they correspond to.

### LINKED ISSUE #934: Chapter 3: System Setup - RStudio IDE Preview version is not yet on Posit website
It's recommended that people use the preview version of RStudio, but it is not yet on the new Posit website. See here: https://community.rstudio.com/t/how-to-find-rstudio-preview-in-the-new-posit-co-site/155376. I'm not sure if it will be available when the book goes to publication so it might be necessary to remove that recommendation?

### PR DESCRIPTION
**Pull request by ateucher**
Closes #791

- Remove IDE preview link and encourage users to update often (closes #934)
- remove devtools history
- included devtools packages in a footnote
- Reframe programmatic vs interactive use of devtools + friends. 
- remove compilation tools section

I was fairly liberal with removing some content, so let me know if I went too far!

### DIFF
--- a/setup.Rmd
+++ b/setup.Rmd
@@ -15,8 +15,8 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
 Make sure you have a recent version of the RStudio integrated development environment (IDE).
 New versions are released regularly, so we recommend updating often to get access to the latest and greatest features.
 
--   Released version: <https://posit.co/download/rstudio-desktop/>
-    -   Most readers can use the free, open source version of RStudio Desktop.
+Download the current version of RStudio Desktop here: <https://posit.co/download/rstudio-desktop/>.
+Most readers can use the free, open source version of RStudio Desktop.
 
 ## devtools, usethis, and you {#sec-setup-usage}
 
@@ -24,38 +24,25 @@ New versions are released regularly, so we recommend updating often to get acces
 >
 > --- Walt Whitman, Song of Myself
 
-After 7 years of development, devtools had grown into a rather unwieldy package, making maintenance difficult.
-Version 2.0.0, released in late 2018, marked the [conscious uncoupling](https://www.tidyverse.org/articles/2018/10/devtools-2-0-0/) of devtools, with most functionality moving into seven smaller packages.
-Through various means, devtools continues to expose all its usual functionality, although it is mostly maintained elsewhere.
+As mentioned in @sec-intro-phil, devtools is a 'meta-package', encompassing and exposing functionality maintained in seven smaller packages[^setup-1].
 For example, devtools might provide a wrapper function in order to set user-friendly defaults, introduce helpful interactive behaviour, or to combine functionality from multiple sub-packages.
+In some cases it simply re-exports a function from another package to make it easily available when devtools is loaded.
+
+[^setup-1]: [remotes](https://remotes.r-lib.org/), [pkgbuild](https://github.com/r-lib/pkgbuild), [pkgload](https://github.com/r-lib/pkgload), [rcmdcheck](https://github.com/r-lib/rcmdcheck), [revdepcheck](https://github.com/r-lib/revdepcheck), [sessioninfo](https://github.com/r-lib/sessioninfo), [usethis](https://usethis.r-lib.org/)
 
 What's our recommended approach to devtools and its constituent packages?
-It varies, depending on whether you're working in useR or developeR mode:
+It varies, depending on your intentions when using the functions:
 
--   For interactive use, useRs should attach devtools and think of it as the provider of your favorite functions for package development.
--   For programmatic use, such as inside another package, developeRs should NOT depend on devtools, but should instead access functions via the package that is their primary home.
-    -   devtools should rarely appear in the role of `pkg` in a qualified call of the form `pkg::fcn()`. Instead, `pkg` should be the package where `fcn()` is defined.
-    -   An exception to this is that we continue to feature `devtools::install_github()` as the way to install the development version of a package in its README, even though `install_github()` actually lives in the remotes package. That's because this piece of advice pertains to interactive use, where we prefer to emphasize devtools.
--   Try to report bugs on the package that is a function's primary home. The help for `devtools::fcn()` usually states when devtools is re-exporting a function from another package.
-
-Example of how to simulate installing and loading a package, during interactive development:
-
-```{r eval = FALSE}
-library(devtools)
-load_all()
-```
-
-If that same functionality is used inside an R package, this is the preferred call:
-
-```{r eval = FALSE}
-pkgload::load_all()
-```
+-   If you are using the functions interactively to help you develop your package, you should think of devtools as the provider of your favorite functions for package development. In this case you should attach devtools and call the functions without qualification (e.g., `load_all()`).
+-   If you are using functions from devtools and friends within the package code you are writing, you should NOT depend on devtools, but should instead access functions via the package that is their primary home.
+    -   devtools should rarely appear in the role of `pkg` in a qualified call of the form `pkg::fcn()`. Instead, `pkg` should be the package where `fcn()` is defined. For example, if you are creating a function in your package in which you need to query the state of the user's R session, use `sessioninfo::session_info()` in your package instead of `devtools::session_info()`.
+-   If you find bugs, try to report them on the package that is a function's primary home. The help for `devtools::fcn()` usually states when devtools is re-exporting a function from another package.
 
 The usethis package is the one constituent package that more people may be aware of and that they may use directly.
-It now holds the functions that act on the files and folders in an R project, most especially for any project that is also an R package.
-All functions in usethis are made available by devtools.
-So, once you attach devtools, you can use any function in usethis without qualification, i.e. just call `use_testthat()`.
-If you choose to specify the namespace, such as when working in a more programmatic style, then access usethis functions directly: do `usethis::use_testthat()` instead of `devtools::use_testthat()`.
+It holds the functions that act on the files and folders in an R project, most especially for any project that is also an R package.
+devtools makes it easy to access usethis functions interactively, as when when you call `library(devtools)`, usethis is also attached.
+Then you can use any function in usethis without qualification, e.g., just call `use_testthat()`.
+If you choose to specify the namespace, such as when working in a more programmatic style, then make sure you qualify the call with usethis, e.g., `usethis::use_testthat()`.
 
 ### Personal startup configuration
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ateucher** [comment]
@jennybc a couple of additional questions for you:

1. Is it worth mentioning why `require()` is used [here](https://github.com/hadley/r-pkgs/blob/6003027de658cba22492e62f192df045216b7a02/setup.Rmd#L76) over `library()`? (because it throws a warning instead of an error if the package isn’t installed, so it won’t interrupt the execution of the rest of the `.Rprofile`?) Or is that an unnecessary tangent?

2. [Here](https://github.com/hadley/r-pkgs/blob/6003027de658cba22492e62f192df045216b7a02/setup.Rmd#L41) it is suggested that `devools::install_github()` is still recommended - I see `remotes::install_github()` very commonly now

File: setup.Rmd [hunk 1]
```diff
@@ -13,52 +13,36 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
...
-In fact, consider using the preview version and updating regularly.
-Compared to the official released version, the preview gives you access to the latest and greatest features and only slightly increases your chances of finding a bug.
-It is distinct from the more volatile daily build.
+New versions are released regularly, so we recommend updating often to get access to the latest and greatest features.
 
--   Preview version: <https://www.rstudio.com/products/rstudio/download/preview/>
--   Released version: <https://www.rstudio.com/products/rstudio/download/>
-    -   Most readers can use the free, open source version of RStudio Desktop.
+Download the current version of RStudio Desktop here: <https://posit.co/download/rstudio-desktop/>.
+Most readers can use the free, open source version of RStudio Desktop.
```
jennybc, line 19: Yeah, this is a good edit. Just to provide you with more background, here's an excerpt from a related slack conversation, when we made similar changes to the pre-workshop setup instructions for WTF. These quotes are from folks in our support organization, confirming my sense that the preview version is much less relevant than in the past:

> ... the Preview release is usually the release candidate for the next general release. It’s not updated often - just before a new release, when the engineering team thinks they have a good candidate. By definition, the Preview release will match the production release until the following release has a candidate, usually a few weeks at most before the next release

> We are definitely releasing more often than we used to, and we therefore put out fewer Previews between GA releases (bigger, longer release cycles mean there’s more risk for regressions and bugs that require new Previews before GA). It’s true that even when Preview and GA differ, there isn’t usually a big gap anymore, with the exception of releases with breaking changes (though those are rare).

> Previews have been utilized less with the change to shorter release cycles. We may have a preview up after "feature complete" for a couple of weeks before a release while we stabilize/certify the final build and that's it.

setup.Rmd
```diff
@@ -13,52 +13,36 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
...
 ## devtools, usethis, and you {#sec-setup-usage}
 
 > "I am large, I contain multitudes."
 >
 > --- Walt Whitman, Song of Myself
 
-After 7 years of development, devtools had grown into a rather unwieldy package, making maintenance difficult.
-Version 2.0.0, released in late 2018, marked the [conscious uncoupling](https://www.tidyverse.org/articles/2018/10/devtools-2-0-0/) of devtools, with most functionality moving into seven smaller packages.
-Through various means, devtools continues to expose all its usual functionality, although it is mostly maintained elsewhere.
+As mentioned in @sec-intro-phil, devtools is a 'meta-package', encompassing and exposing functionality maintained in seven smaller packages[^setup-1].
```
jennybc, line 27:
```suggestion
As mentioned in @sec-intro-phil, devtools is a 'meta-package', encompassing and exposing functionality maintained in several smaller packages[^setup-1].
```

I have learned the hard way that vague wording holds up better 😬

setup.Rmd
```diff
@@ -13,52 +13,36 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
...
 For example, devtools might provide a wrapper function in order to set user-friendly defaults, introduce helpful interactive behaviour, or to combine functionality from multiple sub-packages.
+In some cases it simply re-exports a function from another package to make it easily available when devtools is loaded.
```
jennybc, line 29:
```suggestion
In some cases it simply re-exports a function from another package to make it easily available when devtools is attached.
```

Sort of a nitpick. We do play loose with load vs. attach in some places in the book, but mostly try to be pretty pedantic about it.

setup.Rmd
```diff
@@ -13,52 +13,36 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
...
-    -   An exception to this is that we continue to feature `devtools::install_github()` as the way to install the development version of a package in its README, even though `install_github()` actually lives in the remotes package. That's because this piece of advice pertains to interactive use, where we prefer to emphasize devtools.
--   Try to report bugs on the package that is a function's primary home. The help for `devtools::fcn()` usually states when devtools is re-exporting a function from another package.
-
-Example of how to simulate installing and loading a package, during interactive development:
-
-```{r eval = FALSE}
-library(devtools)
-load_all()
-```
+[^setup-1]: [remotes](https://remotes.r-lib.org/), [pkgbuild](https://github.com/r-lib/pkgbuild), [pkgload](https://github.com/r-lib/pkgload), [rcmdcheck](https://github.com/r-lib/rcmdcheck), [revdepcheck](https://github.com/r-lib/revdepcheck), [sessioninfo](https://github.com/r-lib/sessioninfo), [usethis](https://usethis.r-lib.org/)
```
jennybc, line 31:
```suggestion
[^setup-1]: At the time of writing, devtools exposes functionality from [remotes](https://remotes.r-lib.org/), [pkgbuild](https://github.com/r-lib/pkgbuild), [pkgload](https://github.com/r-lib/pkgload), [rcmdcheck](https://github.com/r-lib/rcmdcheck), [revdepcheck](https://github.com/r-lib/revdepcheck), [sessioninfo](https://github.com/r-lib/sessioninfo), and [usethis](https://usethis.r-lib.org/)
```

More rewording to be a bit more future-proof. Shouldn't this include roxygen2 and testthat?

setup.Rmd
```diff
@@ -13,52 +13,36 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
...
 
-If that same functionality is used inside an R package, this is the preferred call:
+What's our recommended approach to devtools and its constituent packages?
+It varies, depending on your intentions when using the functions:
```
jennybc, line 34:
```suggestion
What's our recommended approach to using devtools and its constituent packages?
It varies, depending on your intention:
```

setup.Rmd
```diff
@@ -13,52 +13,36 @@ install.packages(c("devtools", "roxygen2", "testthat", "knitr"))
...
 
-```{r eval = FALSE}
-pkgload::load_all()
-```
+-   If you are using the functions interactively to help you develop your package, you should think of devtools as the provider of your favorite functions for package development. In this case you should attach devtools and call the functions without qualification (e.g., `load_all()`).
```
jennybc, line 36:
```suggestion
-   If you are using the functions interactively to help you develop your package, you should think of devtools as the provider of your favorite functions for package development. In this case you should attach devtools with `library(devtools)` and call the functions without qualification (e.g., `load_all()`).
```

setup.Rmd
```diff
@@ -109,45 +93,6 @@ devtools::install_github("r-lib/devtools")
 devtools::install_github("r-lib/usethis")
 ```
 
-## R build toolchain {#setup-tools}
```
jennybc, line 112: I'm not sure we can entirely remove this. But in the name of merging this PR, which is overwhelmingly helpful, I can accept the deletion. But will you open an issue about re-inserting some suitable stub re: setting your machine for package development? And paste in the text that we are deleting?

jennybc (commented):
Thanks! This is very helpful.

> Is it worth mentioning why require() is used ...?

This seems like a good footnote. And in that footnote link to the section of the book where we talk about "the answer is *almost never* `require()`". https://r-pkgs.org/dependencies-mindset-background.html#sec-dependencies-attach-vs-load

> Here it is suggested that `devools::install_github()` is still recommended - I see `remotes::install_github()` very commonly now

OMG this is something of a recuring/inside joke now. If I had $5 for every time I've had this conversation, I'd be 🤑. People like to discuss this point.

We really do mean to suggest `devtools::install_github()`, because we think of devtools as more of the user-facing interactive workflow package.

If we change this advice, we would skip straight to recommending pak, not remotes, which is basically superseded.

This is also something I don't care hugely about, so if folks choose to suggest `remotes::install_github()` in their READMEs, I am not going to cowboy up and try to enforce my views on them.

jennybc:
I updated the description to include the magic "closes" wording. As long as we open a new issue to think about the system setup stub, I think this closes that issue. I'm not as convinced as Hadley that complete removal is the way to go.

ateucher:
Sounds good about the `require` footnote - I'll add that. 

> If I had $5 for every time I've had this conversation, I'd be 🤑.

Do you accept e-transfer? ;)

setup.Rmd
ateucher: I'm putting that on the list of "life skills from Jenny" 😜

setup.Rmd
ateucher: Yes, great catch

setup.Rmd
ateucher: Will do - sorry I was a bit overzealous!

setup.Rmd
ateucher: Some of these links are to the GitHub repos, and some are to the pkgdown pages. I got the original list (and links) from the [conscious uncoupling](https://www.tidyverse.org/blog/2018/10/devtools-2-0-0/#conscious-uncoupling) blog, but that was a while ago now. Should they all be pkgdown now? On the [devtools pkgdown site](https://devtools.r-lib.org/#conscious-uncoupling) all of the links go to the repos...

setup.Rmd
ateucher: Opened in #937

ateucher:
@jennybc I think I've addressed your first round of comments (thank you!), and opened the issue to add back in the toolchain setup (#937). I had one more question above for you about links to pkgdown pages vs github repos for the devtools packages.

setup.Rmd
jennybc: Re: why the blog didn't list roxygen2 and testhat. Presumably that's because those two packages were already well-established and had a full life of their own at the time of the uncoupling. Versus being created by carving out functionality from devtools itself.

So I think, given how we're approaching the history here, it's appropriate to include roxygen2 and testthat in this list.

I think we should link to a pkgdown site when it exists. But it won't for some of the more low-level packages, in which case GitHub repo is fine.

jennybc (approved):
OK I answered that one question and, once that's implemented, this is ready to merge. Thanks again!

I approved in anticipation of that, if you have that permission (I'm guessing not). In that case, I will merge.

setup.Rmd
ateucher: Perfect, thanks for that explanation - I'll update the links to pkgdown where available.

ateucher:
I do have the permission to merge! I've updated the urls to what they will be once they are all successfully updated - I think that it's probably ok to merge this now under the assumption those pkgdown sites work as planned?

jennybc:
LGTM! I'm happy for you to merge.

Since we haven't merged this yet, we could go ahead and retain the "R build toolchain" setup here, since you and I have already agreed on that in #937.

We could leave #937 open, but all that's left is the question of whether we should say a bit more for Linux users.

ateucher:
Ok, I'll that back in verbatim and merge - thanks!

Issue #888, opened 2022-11-21
Links to WRE


jennybc:
Devise some standard approach where we can note, for each chapter, the most relevant section of WRE.

Split off and generalized from part of #805.

jennybc:
Obviously that should be a hyperlink. Do it such a way that it's not terribly painful to switch from the official page to our Quarto-ized one (or *vice versa*) for these links.

ateucher:
Unfortunately I think switching between the Quarto-ized version and the official pages will be painful because their structure is quite different even though they're built from a common source (they leverage the nodes and sections/subsections etc differently). I don't think there's an easy way to reliably convert a url from one to the other except for possibly the basic chapter headings.

For example:
These two urls are for the same content on the two pages:

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#An-example
vs
https://rstudio.github.io/r-manuals/r-exts/Creating-R-packages.html#an-example
([source](https://github.com/wch/r-source/blob/8c7dcc5779eb9bdebd9e26a4a0dbd26169c13a6c/doc/manual/R-exts.texi#L4593))

as are these:

https://cran.r-project.org/doc/manuals/r-release/R-exts.html#An-external-pointer-example
vs
https://rstudio.github.io/r-manuals/r-exts/System-and-foreign-language-interfaces.html#an-example
([source](https://github.com/wch/r-source/blob/8c7dcc5779eb9bdebd9e26a4a0dbd26169c13a6c/doc/manual/R-exts.texi#L13925))

jennybc:
I think the right move might be to introduce one level of indirection and maintain a central list of WRE links, i.e. use reference-style links instead of inline. Or do something in the redirect file? Then, at least, those are collected and easy to enforce some sort of system on. But I haven't yet looked into how/if this works in Quarto or not. I figure the next stage of the publishing process will introduce various constraints around URLs and we'll tackle it then.

ateucher:
The redirects file is a good idea, though I wonder how that will work when converted to print?

Unfortunately reference-style links don't appear to work in a multi-file book format (i.e., I don't think we could store the links in a single file and have them be referenced from different files across the book).

Another option, though maybe really ugly, would be to write an R function that you call in an inline `r` chunk that generates a markdown link from a table of links, with an argument for the cran or quarto version of the link?

ateucher:
Ok great. Would it be worth at least making sure they are all to the CRAN pages or all to the quarto pages?

[...]

PR #945, opened 2023-02-22
Update Linux content in System Setup & Structure and State

### LINKED ISSUE #851: Update the `.libPaths()` content

### LINKED ISSUE #937: Do we need more re: build toolchain for compiled code for linux users?
This was removed in #936, but we should probably re-add it. This is the original text:

```
## R build toolchain {#setup-tools}

To be fully capable of building R packages from source, you'll also need a compiler and a few other command line tools.
This may not be strictly necessary until you want to build packages containing C or C++ code.
Especially if you are using RStudio, you can set this aside for now.
The IDE will alert you and provide support once you try to do something that requires you to setup your development environment.
Read on for advice on doing this yourself.

### Windows

On Windows the collection of tools needed for building packages from source is called Rtools.

Rtools is NOT an R package.
It is NOT installed with `install.packages()`.
Instead, download it from <https://cran.r-project.org/bin/windows/Rtools/> and run the installer.

During the Rtools installation you may see a window asking you to "Select Additional Tasks".

-   Do *not* select the box for "Edit the system PATH". devtools and RStudio should put Rtools on the `PATH` automatically when it is needed.
-   Do select the box for "Save version information to registry". It should be selected by default.

### macOS

You need to install the Xcode command line tools, which requires that you [register as an Apple developer](https://developer.apple.com/programs/register/) (don't worry, it's free).

Then, in the shell, do:


    ```shell
    xcode-select --install
    ```

Alternatively, you can install the current release of full [Xcode from the Mac App Store](https://itunes.apple.com/ca/app/xcode/id497799835?mt=12).
This includes a very great deal that you do not need, but it offers the advantage of App Store convenience.

### Linux

Make sure you've installed not only R, but also the R development tools.
For example, on Ubuntu (and Debian) you need to install the `r-base-dev` package.

```

ateucher:
Adds additional information for Linux in:
- setting up a user library (Package Structure and State). Closes #851
- more details on installing Linux toolchain (System Setup). Closes #937 

It's possible I went too far down the `.libPaths()` rabbit hole, so let me know if I should remove some of the more esoteric details.

### DIFF
--- a/structure.Rmd
+++ b/structure.Rmd
@@ -361,16 +361,17 @@ Above, the macOS system is used as a primary development machine and has many pa
 The core set of base and recommended packages that ship with R live in the system-level library and are the same on all operating systems.
 This separation appeals to many developers and makes it easy to, for example, clean out your add-on packages without disturbing your base R installation.
 
-[^structure-3]: For more details, see the [Maintaining R section](https://whattheyforgot.org/maintaining-r.html#how-to-transfer-your-library-when-updating-r) in *What They Forgot To Teach You About R*, [Managing Libraries](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Managing-libraries) in *R Installation and Administration* and the R help files for `?Startup` and `?.libPaths`.
+[^structure-3]: For more details, see the [Maintaining R section](https://whattheyforgot.org/maintaining-r.html#how-to-transfer-your-library-when-updating-r) in *What They Forgot To Teach You About R*, [Managing Libraries](https://rstudio.github.io/r-manuals/r-admin/Add-on-packages.html#managing-libraries) in *R Installation and Administration* and the R help files for `?Startup` and `?.libPaths`.
 
 If you're on macOS or Linux and only see one library, there is no urgent need to change anything.
 But next time you upgrade R, consider creating a user-level library.
 By default, R looks for a user library found at the path stored in the environment variable `R_LIBS_USER`, which itself defaults to `~/Library/R/m/x.y/library,` on macOS, and `~/R/m-library/x.y` on Linux (where `m` is a concise description of your CPU architecture, and `x.y` is the R version).
 You can see this path with `Sys.getenv("R_LIBS_USER")`.
 These directories do not exist by default, and the use of them must be enabled by creating the directory.
-When you install a new version of R, and prior to installing any add-on packages, use `dir.create(Sys.getenv("R_LIBS_USER"), recursive = TRUE)` to set up a user library.
+When you install a new version of R, and prior to installing any add-on packages, use `dir.create(Sys.getenv("R_LIBS_USER"), recursive = TRUE)` to create a user library in the default location.
 Now you will have the library setup seen above.
-Alternatively, you could setup a user library elsewhere and tell R about that by setting the `R_LIBS_USER` environment variable in the `.Renviron` in your `HOME` directory.
+Alternatively, you could set up a user library elsewhere and tell R about that by setting the `R_LIBS_USER` environment variable in `.Renviron`.
+The simplest way to edit your `.Renviron` file is with `usethis::edit_r_environ()`, which will create the file if it doesn't exist, and open it for editing.
 
 The filepaths for these libraries also make it clear they are associated with a specific version of R (4.2.x at the time of writing), which is also typical.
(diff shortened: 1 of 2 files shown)


structure.Rmd
```diff
@@ -355,20 +355,22 @@ In both cases we see two active libraries, consulted in this order:
...
-The core set of base and recommended packages that ship with R live in the system-level library and are the same on macOS and Windows.
+The core set of base and recommended packages that ship with R live in the system-level library and are the same on all operating systems.
 This separation appeals to many developers and makes it easy to, for example, clean out your add-on packages without disturbing your base R installation.
 
-[^structure-3]: For more details, see the [Maintaining R section](https://whattheyforgot.org/maintaining-r.html#how-to-transfer-your-library-when-updating-r) in *What They Forgot To Teach You About R*.
+[^structure-3]: For more details, see the [Maintaining R section](https://whattheyforgot.org/maintaining-r.html#how-to-transfer-your-library-when-updating-r) in *What They Forgot To Teach You About R*, [Managing Libraries](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Managing-libraries) in *R Installation and Administration* and the R help files for `Startup` and `.libPaths()`.
```
jennybc, line 364: I am actively mulling over what to do about links into WRE: link to CRAN's version or our Quarto-ified version (<https://rstudio.github.io/r-manuals/r-exts/>, which I think is much more pleasant to read)? This is being tracked in #888 and we certainly aren't going to decide that here. This is just an FYI.

I think these tweaks are more consistent with what we're doing elsewhere.

```suggestion
[^structure-3]: For more details, see the [Maintaining R section](https://whattheyforgot.org/maintaining-r.html#how-to-transfer-your-library-when-updating-r) in *What They Forgot To Teach You About R*, [Managing Libraries](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Managing-libraries) in *R Installation and Administration* and the R help files for `?Startup` and `?.libPaths`.
```

structure.Rmd
```diff
@@ -355,20 +355,22 @@ In both cases we see two active libraries, consulted in this order:
...
+By default, R looks for a user library found at the path stored in the environment variable `R_LIBS_USER`, which itself defaults to `~/Library/R/m/x.y/library,` on macOS, and `~/R/m-library/x.y` on Linux (where `m` is a concise description of your CPU architecture, and `x.y` is the R version).
+You can see this path with `Sys.getenv("R_LIBS_USER")`.
+These directories do not exist by default, and the use of them must be enabled by creating the directory.
 When you install a new version of R, and prior to installing any add-on packages, use `dir.create(Sys.getenv("R_LIBS_USER"), recursive = TRUE)` to set up a user library.
 Now you will have the library setup seen above.
-Alternatively, you could setup a user library elsewhere and tell R about that by setting the `R_LIBS_USER` environment variable in `.Renviron`.
+Alternatively, you could setup a user library elsewhere and tell R about that by setting the `R_LIBS_USER` environment variable in the `.Renviron` in your `HOME` directory.
```
jennybc, line 373: Let's not get into where `.Renviron` is looked for, because technically it's not quite this simple. We can just remind folks to use `usethis::edit_r_environ()` to open it for editing.

jennybc (approved):
Thanks! I think my suggestions are fairly modest, so I'm pre-approving so you can merge once you're content.

structure.Rmd
ateucher: Ah, great idea.

structure.Rmd
ateucher: Yes, the quarto version of WRE is much nicer to read - though this link is not to WRE rather R administration.

jennybc: Ah, I see. We Quarto-ized all the manuals, so it'll be the same decision:

https://rstudio.github.io/r-manuals/r-admin/

ateucher: Oh I didn't realize that - lovely!

jennybc:
Shall we merge this?

PR #951, opened 2023-03-02
Include VS Code as alternative to RStudio

### LINKED ISSUE #949: Consider mentioning vscode?
Visual Studio Code wasn't really on the landscape when the first edition of the book was released, but it's now obviously a major player. I feel like VSCode should be mentioned [alongside Emacs + ESS & Vim here](https://github.com/hadley/r-pkgs/blob/d5c9b45a9e1f896c684061516e863b78a03a16d5/introduction.Rmd#L61) (and/or consider whether or not Emacs and Vim are still in wide enough use with the target audience of the book - I don't have a good feel for that)?

ateucher:
Include VS Code, move ESS & vim to footnote. Closes #949


jennybc (approved):
Thanks, LGTM!

PR #957, opened 2023-03-14
Update testing-design.Rmd

jadeynryan:
correct minor typo


jennybc:
Thanks! It turns out I fixed this recently already.

PR #962, opened 2023-03-22
Update man.Rmd

jadeynryan:
Correct minor typos (elided to eluded; extra `}`)


jennybc:
Thanks for alerting me to the stray bracket. I removed that.

As for "elided", that really is the word I mean to use.

PR #976, opened 2023-04-01
Update software-development-practices.Rmd

matanhakim:
Maybe this is supposed to be `use_github_actions()` rather than`use_github_action()`? When I type `use_github_action()` I get:
```
> use_github_action()
Error in check_string(name) : argument "name" is missing, with no default
```


jennybc:
No, the promised behaviour is only in the dev version of usethis, which I'll release soon.

PR #978, opened 2023-04-03
wre links

ateucher:
I don't know if this is still useful at this point, but I changed all of the links to WRE and R-Admin manuals to the quarto-ized versions. I also included a csv lookup of the links that are used with urls to the quarto version and the cran version.


jennybc:
Once the print version is completely out of my hair, I will take these changes.

PR #981, opened 2023-04-16
Typos chap3

petzi53:


jennybc:
Thanks! This material is in a weird state at the moment, as things are sort of "frozen" while I work through the process of publishing the printed version. As such, I'm really only taking changes that are unequivocally errors/typos. I had already found and fixed the first thing you report and I consider the other two to be sort of stylistic. So I'm closing this unmerged, but thanks again.

PR #988, opened 2023-05-02
Clarify a sentence on methods

howardbaek:
I was confused about this sentence, _For instance, because R’s main object-oriented systems (S3 and S4) are based on generic functions (i.e., methods belong to functions not classes), testing approaches built around objects and methods don’t make much sense._ 

So, I watched Roger Peng's "Classes and Methods in R" lecture to clarify things.

Slide taken from https://youtu.be/93N0HdoZW9g?t=364

Screenshot 2023-05-02 at 10 27 16 AM

Maybe we can use the last bullet point on this slide to replace what's in the parenthesis?

### DIFF
--- a/testing-basics.Rmd
+++ b/testing-basics.Rmd
@@ -71,7 +71,7 @@ This chapter describes how to test your R package using the testthat package: <h
 
 If you're familiar with frameworks for unit testing in other languages, you should note that there are some fundamental differences with testthat.
 This is because R is, at heart, more a functional programming language than an object-oriented programming language.
-For instance, because R's main object-oriented systems (S3 and S4) are based on generic functions (i.e., methods belong to functions not classes), testing approaches built around objects and methods don't make much sense.
+For instance, because R's main object-oriented systems (S3 and S4) are based on generic functions (i.e., methods belong to generic functions not classes), testing approaches built around objects and methods don't make much sense.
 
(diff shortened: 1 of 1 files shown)


testing-basics.Rmd
```diff
@@ -71,7 +71,7 @@ This chapter describes how to test your R package using the testthat package: <h
 
 If you're familiar with frameworks for unit testing in other languages, you should note that there are some fundamental differences with testthat.
 This is because R is, at heart, more a functional programming language than an object-oriented programming language.
-For instance, because R's main object-oriented systems (S3 and S4) are based on generic functions (i.e., methods belong to functions not classes), testing approaches built around objects and methods don't make much sense.
+For instance, because R's main object-oriented systems (S3 and S4) are based on generic functions (i.e., a method is the implementation of a generic function for an object of a particular class), testing approaches built around objects and methods don't make much sense.
```
jennybc, line 74: If we're going to touch this, I think it needs to be tweaked.
```suggestion
For instance, because R's main object-oriented systems (S3 and S4) are based on generic functions (i.e., a method implements a generic function for a specific class), testing approaches built around objects and methods don't make much sense.
```

jennybc:
Thanks!

howardbaek:
Thanks for your work on this book! Its super helpful.

PR #1012, opened 2023-07-23
typo fix

ktang:


jennybc:
This is intentional. The main conference for the R language is called `useR!` and people sometimes use useR to refer to people who use R.

ktang:
Thank you for the explanation!

PR #1022, opened 2023-09-19
Change in windows user library location (for R>=4.2)

mtkerbeR:
@jennybc 
As discussed in [rstats-wtf](https://github.com/rstats-wtf/what-they-forgot/pull/98), standard for windows user library has changed starting with R4.2, see [PR#17842](https://bugs.r-project.org/show_bug.cgi?id=17842)

Update for path to windows user library, which is (for R>=4.2) located at C:/Users/<username>/AppData/Local/R/win-library/x.y


jennybc:
Thanks!

PR #1025, opened 2023-10-04
fix function: expect_equal_tbls, in section 14.3.1

orderlyquant:
I believe the changes made reflect the author's intention.

"I assign the copyright of this contribution to Hadley Wickham".


jennybc:
Things are the way they are because we truly did copy live code from dbplyr at a specific point in time and `rest` is what that uses. But, yeah, I think given our redaction, it would be fine and slightly better to refer to `results`.

PR #1027, opened 2023-10-31
Clarify that internal data is always lazy-loaded

salim-b:
Cf. https://stackoverflow.com/questions/41269434/r-package-make-internal-data-not-lazy

### DIFF
--- a/data.Rmd
+++ b/data.Rmd
@@ -235,7 +235,7 @@ But you can surface it with `check(cran = FALSE, env_vars = c("_R_CHECK_PACKAGE_
 Sometimes your package functions need access to pre-computed data.
 If you put these objects in `data/`, they'll also be available to package users, which is not appropriate.
 Sometimes the objects you need are small and simple enough that you can define them with `c()` or `data.frame()` in the code below `R/`, perhaps in `R/data.R`.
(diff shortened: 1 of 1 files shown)


jennybc:
Interesting! I'm happy to merge but would feel even better if we had a reference that was more official, e.g. from the official R manuals. In the journey that brought you here, have you encountered any explicit statement about the laziness of internal data in official R docs?

salim-b:
The only official information I could find is the R Manual which [states](https://rstudio.github.io/r-manuals/r-exts/Creating-R-packages.html#package-subdirectories):

> Two exceptions are allowed: if the R subdirectory contains **a file `sysdata.rda`** (a saved image of one or more R objects: please use suitable compression as suggested by `tools::resaveRdaFiles`, and see also the `SysDataCompression` `DESCRIPTION` field.) this **will be lazy-loaded** into the namespace environment – this is intended for system datasets that are not intended to be user-accessible via data.

It doesn't *explicitly say* "always". But since no known means to disable lazy-loading of internal data seem to exist, it *practically means* "always", I guess.

jennybc:
OK that's reassuring to find. Thanks!

PR #1035, opened 2023-11-28
Rendering Issues

davidrsch:
These commits fixes some issues in rendering process

### DIFF
--- a/common.R
+++ b/common.R
@@ -106,3 +106,5 @@ status <- function(type) {
     ":::\n"
   ))
 }
+
+options(repos = "http://cran.us.r-project.org ")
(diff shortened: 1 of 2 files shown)


jennybc:
Can you say more? Like, "if you look at X, there's this obvious problem, so Y fixes it". Not that these are huge changes, but I also spent lots of time with this content quite recently and did not notice problems that require these changes.

davidrsch:
Yes, sorry, I would try to explain more. I tried to render the book using my PC and while rendering the book I received those error messages. I am currently using R.4.3.1 and have all the packages necessaries for the book upto dated via CRAN. I use a windows. If no one else got these error messages could be a local issue (from my PC or packages versions perhaps).

jennybc:
I don't want to hardwire a specific CRAN mirror in the book. I think most folks who would try to render the book locally have external arrangements for setting their CRAN mirror that apply to all of their R work.

I also don't understand the need for `error = TRUE` in that chunk from the testing chapter.

So, yes, I think I am going to leave this unmerged, as I'm not seeing evidence locally or in CI that these are needed.

PR #1038, opened 2023-12-13
Update lifecycle.Rmd

MikeJohnPage:
In [section 21.2](https://r-pkgs.org/lifecycle.html#sec-lifecycle-version-number) it states that the function `package_version()` comes from utils, but I think this is a mistake and is supposed to say that it comes from base.

I've updated the sections of relevant text to reflect this, but now I am not sure the updated footnote indicating that base should be added to imports (previously it said utils should be added) is relevant/required.

### DIFF
--- a/lifecycle.Rmd
+++ b/lifecycle.Rmd
@@ -96,7 +96,7 @@ These three different views of usethis's evolution are all useful for different 
 
 Formally, an R package version is a sequence of at least two integers separated by either `.` or `-`.
 For example, `1.0` and `0.9.1-10` are valid versions, but `1` and `1.0-devel` are not.
-Base R offers the `base::package_version()`[^lifecycle-2] function to parse a package version string into a proper S3 class by the same name.
+Base R offers the `package_version()` function to parse a package version string into a proper S3 class by the same name.
 This class makes it easier to do things like compare versions.
 
 [^lifecycle-2]: We can call `package_version()` directly here, but in package code, you should use the `base::package_version()` form and list the base package in `Imports`.
--- a/lifecycle.Rmd
+++ b/lifecycle.Rmd
@@ -99,8 +99,6 @@ For example, `1.0` and `0.9.1-10` are valid versions, but `1` and `1.0-devel` ar
 Base R offers the `package_version()` function to parse a package version string into a proper S3 class by the same name.
 This class makes it easier to do things like compare versions.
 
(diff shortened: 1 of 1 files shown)


jennybc:
Wow, I'm not sure how I did this! At first I thought maybe `package_version()` moved from utils to base?!? But of course it has not. In any case, thanks for the catch.

(Confusingly, there is `utils::packageVersion()` but that's not what I'm talking about in this section of the book.)

Yeah, the proposed fix isn't quite right yet, because `package_version()`'s presence in base, as opposed to utils, negates most of what's being said. Now that I've confirmed your hunch, you can take another crack. Or next time I come through I'll do it.

lifecycle.Rmd
```diff
@@ -96,10 +96,10 @@ These three different views of usethis's evolution are all useful for different
 
 Formally, an R package version is a sequence of at least two integers separated by either `.` or `-`.
 For example, `1.0` and `0.9.1-10` are valid versions, but `1` and `1.0-devel` are not.
-Base R offers the `utils::package_version()`[^lifecycle-2] function to parse a package version string into a proper S3 class by the same name.
+Base R offers the `base::package_version()`[^lifecycle-2] function to parse a package version string into a proper S3 class by the same name.
```
jennybc, line 99:
```suggestion
Base R offers the `package_version()` function to parse a package version string into a proper S3 class by the same name.
```

lifecycle.Rmd
```diff
@@ -96,10 +96,10 @@ These three different views of usethis's evolution are all useful for different
...
 This class makes it easier to do things like compare versions.
 
-[^lifecycle-2]: We can call `package_version()` directly here, but in package code, you should use the `utils::package_version()` form and list the utils package in `Imports`.
+[^lifecycle-2]: We can call `package_version()` directly here, but in package code, you should use the `base::package_version()` form and list the base package in `Imports`.
```
jennybc, line 102: This should just go away.

MikeJohnPage: Agreed

MikeJohnPage (commented):
Footnote has been removed and your suggest changes committed.

MikeJohnPage:
> (Confusingly, there is utils::packageVersion() but that's not what I'm talking about in this section of the book.)

Yes, this stumped me too. For a while I was convinced it was just a mistake of using snake_case instead of camelCase, but then I clicked they are different functions!

It should be okay to merge now I think. Thanks.

MikeJohnPage:
Also, this book is great, so thanks for all your hard work on it - it is appreciated 👏

PR #1040, opened 2024-01-17
Update advice to use `use_standalone()`

olivroy:


dependencies-in-practice.Rmd
```diff
@@ -776,7 +776,12 @@ Here is a redacted version of glue's `.onLoad()` function, where you'll see that
 ```
 
 The `s3_register()` function comes from the vctrs package.
-If you don't have an organic need to depend on vctrs, it is common (and encouraged) to simply inline [the `s3_register()` source](https://github.com/r-lib/vctrs/blob/main/R/register-s3.R) into your own package.
+If you don't have an organic need to depend on vctrs, it is common (and encouraged) to simply inline [the `s3_register()` source](https://github.com/r-lib/vctrs/blob/main/R/register-s3.R) into your own package with
```
jennybc, line 779:
```suggestion
If you don't have an organic need to depend on vctrs, it is common (and encouraged) to simply inline [the `s3_register()` source](https://github.com/r-lib/vctrs/blob/main/R/register-s3.R) into your own package.
One way to do this is with `usethis::use_standalone()`:
```

jennybc (commented):
Thanks!

I've proposed a rephrasing that feels better given it feels like we haven't really covered `use_standalone()` in the book yet. Assuming we eventually do address it specifically somewhere, someday, this would become a crosslink to that.

olivroy:
You are right. It is not introduced anywhere in the book. But I think it may be worth to refer to it. As it can be a good excuse to talk about other standalone files that can be used.

jennybc:
Yeah, `use_standalone()` just matured after 2e was basically "done".

jennybc:
The CI failure is because P3M is having a bad day.

PR #1043, opened 2024-02-15
properly enclose function in back ticks

olivroy:


jennybc:
Thanks! Ah, visual editor woes 😬

olivroy:
I think #1049 is a duplicate.

FWIW, I tried locally switching back and forth with the visual editor. Seems like the back ticks are no longer wrongly escaped. It has possibly improved since then!

jennybc:
Thanks for spotting the duplicate. I'll take the earlier PR.

Closes #1049.

PR #1045, opened 2024-03-06
Update introduction.Rmd

Fgazzelloni:
typo


jennybc:
What we have currently is most consistent with the tidyverse style guide re: capitalization:

https://style.tidyverse.org/documentation.html?q=package#capitalization-and-full-stops

so I will leave this for now. Mostly we try to avoid putting a package name at the beginning of a sentence, because it feels like no capitalization choice is correct. But apparently I couldn't find a way to write this sentence without putting "devtools" first.

Fgazzelloni:
Another option is to start with something else than the name of the package direclty. Such as: The {devtools} ...bla bla

PR #1046, opened 2024-03-07
Update other-markdown.Rmd

Fgazzelloni:
consider using `use_github_action()` instead.


jennybc:
Thanks! I'm going to merge a different PR that addresses this, because it has another change that we also need.

PR #1050, opened 2024-03-18
Rename index.Rmd to 00-welcome.Rmd

gungorMetehan:
renaming chapters for a well-organized folder (suggestion)


jennybc:
I will assume this PR (and its 35 friends) are well-intentioned, but I'm afraid this is not helpful. 

We are OK with our current filenames and even if we wanted to change them, it's a relatively easy bulk change for us. The fact that it's 35 separate PRs and also doesn't include the additional changes necessary in `_quarto.yml` for the book to actually build makes this feel like an effort to juice one's GitHub activity stats.

PR #1085, opened 2024-08-09
Remove duplicate word 'section' in ch 7

cgeger:
removes duplicate word 'section' from text as read aloud (e.g. reads 'Section 7.2' instead of 'section Section 7.2'


jennybc:
Thanks!

PR #1087, opened 2024-08-15
describe @exportS3Method for non-owned generics and delayed method registration

### LINKED ISSUE #1032: Update S3 method advice
To use `@exportS3Method` instead of `rlang::s3_register()`.

### LINKED ISSUE #1082: Update the section about conditional / delayed S3 registration
This section:

https://r-pkgs.org/dependencies-in-practice.html#imports-and-exports-related-to-s3

needs to be updated to reflect a cleaner solution that is available from R >= 3.6 and that is newly taken advantage of by roxygen2.

A pull request showing before vs. after is here:

https://github.com/tidyverse/glue/pull/310

tanho63:
Closes #1032 and closes #1082. 

Cleans up the section about exporting S3 methods for non-owned generics to recommend using `exportS3Method` in all scenarios (hard and soft dependencies). 

- For generics owned by hard dependencies (ie `Depends` and `Imports`), this previously described a workflow that required importing the generic into the NAMESPACE so that the method could be exported via `@export` - it seems like this can be skipped and we can use `@exportS3Method` directly in this scenario.  

- For generics owned by soft dependencies (ie `Suggests`), this described a methodology around registering S3 methods `onLoad()`, and suggested an `s3_register()` function that needed to be included in the package body. This is no longer required as of R 3.6, which now interprets the `S3method()` directive in NAMESPACE to automatically register the method on dependency load. This means we can use `@exportS3Method` for this scenario as well, and remove the section on conditional/delayed load entirely. 

I'm unsure if it would be worth mentioning how to handle things for pre R3.6 (and hide the deleted section in a collapsed details block?) - R 3.6 is quite old now.

### DIFF
--- a/dependencies-in-practice.Rmd
+++ b/dependencies-in-practice.Rmd
@@ -656,7 +656,7 @@ count <- function(x, ..., wt = NULL, sort = FALSE, name = NULL) {
   UseMethod("count")
 }
 
-#' @exportS3method
+#' @export
 count.data.frame <- function(
   x,
   ...,
--- a/dependencies-in-practice.Rmd
+++ b/dependencies-in-practice.Rmd
@@ -679,7 +679,7 @@ export(count)
 Now imagine that your package implements a method for `count()` for a class you "own" (not `data.frame`).
 A good example is the dbplyr package, which implements `count()` for the `tbl_lazy` class.
 
-In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`, providing the precse generic that we're providing a method for.
+In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`, providing the precise generic that we're providing a method for.
 This will use "conditional" method registration, which means the method will only be registered when the dplyr package is loaded.
 
 ```{r}
--- a/dependencies-in-practice.Rmd
+++ b/dependencies-in-practice.Rmd
@@ -690,7 +690,7 @@ count.tbl_lazy <- function(x, ..., wt = NULL, sort = FALSE, name = NULL) { ... }
 In `NAMESPACE`, we have:
 
 ```         
-S3method(count,tbl_lazy)
+S3method(dplyr::count,tbl_lazy)
 ```
 
 This also works for generics for packages that are suggested dependencies, e.g. where the glue package implements a method for testthat's `compare()` generic: glue lists testthat as only a suggested dependency, so (as of R 3.6.0) R will conditionally register the method `compare.glue()` when both the testthat and the glue packages are loaded.
--- a/dependencies-in-practice.Rmd
+++ b/dependencies-in-practice.Rmd
@@ -679,8 +679,8 @@ export(count)
 Now imagine that your package implements a method for `count()` for a class you "own" (not `data.frame`).
 A good example is the dbplyr package, which implements `count()` for the `tbl_lazy` class.
 
-In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use @exportS3Method.
-This will use "delayed" method registration, which means the method will only be registered when the dplyr package is loaded.
+In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`.
+This will use "conditional" method registration, which means the method will only be registered when the dplyr package is loaded.
 
 ```{r}
 #| eval: false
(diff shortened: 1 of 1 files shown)


dependencies-in-practice.Rmd
```diff
@@ -656,7 +656,7 @@ count <- function(x, ..., wt = NULL, sort = FALSE, name = NULL) {
   UseMethod("count")
 }
 
-#' @export
+#' @exportS3method
```
hadley, line 659: I think that this can still be `@export`? I'm not sure I'm yet ready to say that every S3 method should use `@exportS3method`.

dependencies-in-practice.Rmd
```diff
@@ -678,35 +678,24 @@ export(count)
...
-
-```         
-Imports: 
-    ...,
-    dplyr,
-    ...
-```
 
-In `dbplyr/R/verb-count.R`, we have:
+In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`.
```
hadley, line 682:
```suggestion
In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`, providing the precse generic that we're providing a method for.
```

dependencies-in-practice.Rmd
```diff
@@ -678,35 +678,24 @@ export(count)
...
 
 ```         
 S3method(count,tbl_lazy)
-...
-importFrom(dplyr,count)
 ```
 
-Dbplyr also provides methods for various generics provided by the base package, such as `dim()` and `names()`.
-In this case, there is no need to import those generics, but it's still necessary to export the methods.
+dbplyr also provides methods for various generics provided by the base package, such as `dim()` and `names()` - in this case, we can use `@export` or `@exportS3Method base::names`
```
hadley, line 697: I think I'd still recommend using `@export` here because we don't need conditional registration.

dependencies-in-practice.Rmd
```diff
@@ -678,35 +678,24 @@ export(count)
...
-```         
-Imports: 
-    ...,
-    dplyr,
-    ...
-```
 
-In `dbplyr/R/verb-count.R`, we have:
+In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`.
+This will use "conditional" method registration, which means the method will only be registered when the dplyr package is loaded.
```
hadley, line 683: I think it would be worth a sentence or two here (or maybe below) to also mention that this also works even if the package is only suggested.

hadley (commented):
Nice improvement! I've added a few comments/questions.

dependencies-in-practice.Rmd
tanho63:
```suggestion
#' @export
```

dependencies-in-practice.Rmd
tanho63:
```suggestion
dbplyr also provides methods for various generics provided by the base package, such as `dim()` and `names()` - in this case, we can still use `@export` since we don't need conditional registration. 
```

dependencies-in-practice.Rmd
```diff
@@ -716,7 +705,7 @@ dim.tbl_lazy <- function(x) {
   c(NA, length(op_vars(x$lazy_query)))
 }
 
-#' @export
+#' @exportS3Method base::names
```
tanho63, line 708:
```suggestion
#' @export
```

dependencies-in-practice.Rmd
```diff
@@ -727,62 +716,5 @@ In `NAMESPACE`, this produces:
...
-```{r}
-#| eval: false
-.onLoad <- function(...) {
-  s3_register("testthat::compare", "glue")
-  s3_register("waldo::compare_proxy", "glue")
-  s3_register("vctrs::vec_ptype2", "glue.glue")
-  ...
-  invisible()
-}
+S3method(base::names,tbl_lazy)
```
tanho63, line 719:
```suggestion
S3method(names,tbl_lazy)
```

dependencies-in-practice.Rmd
```diff
@@ -678,35 +678,24 @@ export(count)
...
-
-```         
-Imports: 
-    ...,
-    dplyr,
-    ...
-```
 
-In `dbplyr/R/verb-count.R`, we have:
+In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`, providing the precse generic that we're providing a method for.
```
tanho63, line 682:
```suggestion
In this case, `@export` will not work because it assumes the `count()` generic is included or imported in the dbplyr NAMESPACE, so instead we need to use `@exportS3Method`, providing the precise generic that we're providing a method for.
```

dependencies-in-practice.Rmd
tanho63: Updated with a sentence below (and removes mention of "conditional" method registration for import dependencies)

dependencies-in-practice.Rmd
```diff
@@ -676,37 +676,27 @@ export(count)
...
-#' @importFrom dplyr count
-#' @export
+#' @exportS3Method dplyr::count
 count.tbl_lazy <- function(x, ..., wt = NULL, sort = FALSE, name = NULL) { ... }
 ```
 
 In `NAMESPACE`, we have:
 
 ```         
 S3method(count,tbl_lazy)
```
hadley, line 693: Ooh this will change too.

tanho63: Whoops, yep

hadley:
Thank you!

PR #1099, opened 2025-05-15
fix typo url

JosephBARBIERDARNAL:
Hi! 

I was browsing this page: https://r-pkgs.org/R-CMD-check.html

And found that this link at the very top https://r-pkgs.org/r-cmd-check.html is broken because of lowercasing.


jennybc:
Thanks!

PR #1119, opened 2026-06-04
Fix dead link in preface

victorhartman:
Link in the preface points to page that no longer exists. Fixed the link.


jennybc:
Thanks!

